PR #7, opened 2019-11-10
do not stop ctrl-c in cargo-bpf load

mockersf:


alessandrod:
Thanks for the patch!

We do want to handle SIGINT tho. If we don't, the process will just be aborted and we'll potentially leave XDP programs installed.

alessandrod:
I have implemented what I think is the correct behaviour here: https://github.com/alessandrod/redbpf/commit/4ad11eb208b7a0dcd78d8933d376fd3a7a89f22e

rsdy:
If we just allow cargo-bpf to be killed, then we need to manually unload the XDP programs, which can be quite annoying. Thanks, but I won't merge this. Enjoy to Pi, though :)

PR #14, opened 2019-11-15
Relicense under Apache2 + MIT

rsdy:
Relicense the project under Apache2+MIT shared, to work better with the rest of the Rust ecosystem.

The README explicitly lists all foreign code, and their licenses, and we should stick with this book keeping in the future.

I've added a file called `PREAMBLE` which should be used as the standard preamble for new files.


rahulpowar:
Looks good to me

PR #22, opened 2020-01-08
Rewrite unsound attribute offset checks

HeroicKatora:
Dereferencing a null pointer into a reference is always unsound. Using a
null pointer and even other techniques to determine attribute offsets
has already cause issues even in crates that are fully dedicated to it.

The result of this are that these tests might in the future unexpectedly fail 
or, worse, unexpectedly and wrongly succeed. As will all UB, it would be 
within the compiler's rights to assume that all `asserts!` hold even when
the actual offsets differ.

However, we are fortunate here as the structs in question are crafted
out of only basic integer types. Thus, it is safe to create an instances
of them that is fully zeroed with little code. Then the actual reference
to member access is fully safe and sound. An exception are only the
fields of unions which are inherently unsafe to access and using them to
create a reference is no exception.

### DIFF
--- a/redbpf/src/sys/perf.rs
+++ b/redbpf/src/sys/perf.rs
@@ -271,8 +271,9 @@ fn bindgen_test_layout___kernel_fd_set() {
         8usize,
         concat!("Alignment of ", stringify!(__kernel_fd_set))
     );
+    let instance = unsafe { core::mem::zeroed::<__kernel_fd_set>() };
     assert_eq!(
-        unsafe { &(*(::std::ptr::null::<__kernel_fd_set>())).fds_bits as *const _ as usize },
+        &instance.fds_bits as *const _ as usize - &instance as *const _ as usize,
         0usize,
         concat!(
             "Offset of field: ",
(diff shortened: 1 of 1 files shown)


rsdy:
Thanks for the patches, it's good to see that soundness issues are a hot topic, and people are keen on improving the situation.

However, in this scenario, I would expect the compiler to do the correct thing and resolve the offset at compile time, so I don't think this changeset makes a material difference in semantics.

Moreover, as noted in the file, this is generated code, so the right place to raise this issue is in [`bindgen`](https://github.com/rust-lang/rust-bindgen). I would prefer not to deviate from `bindgen` generated code, but I'm happy to merge a PR once this change is accepted there.

HeroicKatora:
> However, in this scenario, I would expect the compiler to do the correct thing and resolve the offset at compile time, so I don't think this changeset makes a material difference in semantics.

The offset resolving at compile time *can* also be anything else. But thanks for point me to `bindgen` for this because that means I have an actionable target in `rust-lang` :smile:

PR #24, opened 2020-01-08
Hide the unsafe to_str function

HeroicKatora:
The implementation makes the assumption that the bytes making up the
C-string are both terminated and valid. This is wrong in general but
the zero-termination seems permissible as it only trusts the kernel's
correct implementation of POSIX.

The correct encoding trust is a different beast, there might be an
unterminated multi-byte character within the string. Also, the implicit trust 
on correct encoding of kernel files is present in a few other places as well
and might warrant a separate tracking issue if it should be addressed.

### DIFF
--- a/bpf-sys/src/uname.rs
+++ b/bpf-sys/src/uname.rs
@@ -49,7 +49,9 @@ pub fn get_fqdn() -> Result<String, ()> {
 }
 
 #[inline]
-pub fn to_str(bytes: &[c_char]) -> &str {
+fn to_str(bytes: &[c_char]) -> &str {
+    // SAFETY: only called on `uname` structs filled by the OS, which we trust.
+    // FIXME: Do not trust it to be valid UTF-8.
     unsafe { from_utf8_unchecked(CStr::from_ptr(bytes.as_ptr()).to_bytes()) }
 }
 
--- a/bpf-sys/src/uname.rs
+++ b/bpf-sys/src/uname.rs
@@ -51,7 +51,6 @@ pub fn get_fqdn() -> Result<String, ()> {
 #[inline]
 fn to_str(bytes: &[c_char]) -> &str {
     // SAFETY: only called on `uname` structs filled by the OS, which we trust.
-    // FIXME: Do not trust it to be valid UTF-8.
     unsafe { from_utf8_unchecked(CStr::from_ptr(bytes.as_ptr()).to_bytes()) }
 }
 


bpf-sys/src/uname.rs
```diff
@@ -49,7 +49,9 @@ pub fn get_fqdn() -> Result<String, ()> {
 }
 
 #[inline]
-pub fn to_str(bytes: &[c_char]) -> &str {
+fn to_str(bytes: &[c_char]) -> &str {
+    // SAFETY: only called on `uname` structs filled by the OS, which we trust.
+    // FIXME: Do not trust it to be valid UTF-8.
```
rsdy: I don't think it's reasonable to expect Linux to return anything other than ASCII in uname. If that's the case, I'd like to see a demonstration of it, and then we can come up with a more generic fix. Please remove this line before merge.

HeroicKatora: The proof obligation is usually on the writer of `unsafe` that the code inside is well behaved. If you can give me a source for the claim that this is true, then I will gradly replace it with a reference to such. The POSIX standard gives no indication that it requires such thing. Otherwise, it might not be much overhead to do a checked conversion instead.

rsdy: I did not try to imply you should come up with a proof of concept. I'm quite happy to keep it this way until an actual problem comes up. That said, I don't think this comment is valid, and since we're going on hunches, I'd like the comment to be gone before merge.

HeroicKatora: > I did not try to imply you should come up with a proof of concept.

But you should have. Adding `unsafe` is best understood as a promise that the programmer has a *proof* that the compiler does not understand. Such a proof does exist for null-termination, in the POSIX docs. In the to-`str` conversion though, especially since there are safe alternatives, the only safe approach would require giving a reason *why* to run unproven `unsafe` code. Has there been a measure performance impact in mission critical code? If so, maybe offer the supposedly more efficient version as an `unsafe` alternative but in general use the safe, checked version? My motivation for including that FIXME was getting an answer to those questions.

>  I'm quite happy to keep it this way until an actual problem comes up.

Adding to the above, this is not how UB works. There will most likely not be any obvious problems that come up should the assumption ever not hold. Most likely it will be mostly invisible and *not* crash anything, just silently behave slightly wrong and then in a truly worst or malicious case snowball into breaking other assumptions.

rsdy:
Reverted.

PR #25, opened 2020-01-09
Enable CircleCI builds on all branches

rsdy:


rsdy:
Instead going with #26

PR #27, opened 2020-01-09
Fix length assertions for data reading

HeroicKatora:
As explained [in this comment][1], the `pointer::add` function comes with its own preconditions and it is unsound to pass it an arbitrary length that has not been validated (because we are in the process of validating it). The fixed version is not only sounder and safer but also cleaner as it reuses encapsulated unsafety which avoids duplicating justifications.

This does not remove all usage of this patterns but might be a usable template for doing so.

[1]: https://github.com/redsift/redbpf/issues/20#issuecomment-572722967


rsdy:
The code generated by this PR will not pass the eBPF verifier, as it expects explicit runtime boundary checks against specific registers which must not be optimised away by the compiler.

PR #30, opened 2020-01-27
Add support for panic!()

alessandrod:
Before this change, code that *potentially* called panic!() would fail to link, as core::panicking::{panic,panic_bound_check,panic_fmt} were not being inlined. That meant that even indexing arrays, unless bounds checks were somehow being elided, could result in compilation failures.

This change introduces a LLVM IR transform pass which among other things forces core::panicking::* to be inlined, and patches the panic handler to use the BPF exit opcode to terminate execution.

If nothing calls panic!(), no extra code gets emitted. If something does panic, you now get a log line in /sys/kernel/debug/tracing/trace_pipe saying "panic in [filename of the BPF program that panicked].rs".

### DIFF
--- a/cargo-bpf/src/build.rs
+++ b/cargo-bpf/src/build.rs
@@ -5,6 +5,7 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+use libc;
 use std::convert::From;
 use std::env;
 use std::fmt::{self, Display};
@@ -12,8 +13,11 @@ use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::slice;
+use std::str;
 use toml_edit;
 
+use crate::llvm::process_ir;
 use crate::CommandError;
 
 #[derive(Debug)]
@@ -21,7 +25,8 @@ pub enum Error {
     MissingManifest(PathBuf),
     NoPrograms,
     NoLLC,
-    Compile(String),
+    NoOPT,
+    Compile(String, Option<String>),
     MissingBitcode(String),
     Link(String),
     IOError(io::Error),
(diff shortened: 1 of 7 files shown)


cargo-bpf/src/build.rs
```diff
@@ -121,6 +150,23 @@ pub fn build_program(
     Ok(())
 }
 
+fn get_opt_executable() -> Result<String, Error> {
+    for llc in vec!["opt".into(), env::var("OPT").unwrap_or("opt-9".into())].drain(..) {
```
rsdy: This copy-pasta from the function below, we should probably keep one LLVM version check.

cargo-bpf/src/llvm.rs
```diff
@@ -0,0 +1,121 @@
...
+                        ptr::null_mut(),
+                        0,
+                        CString::new("").unwrap().as_ptr(),
+                    );
+                }
+            }
+            func = LLVMGetNextFunction(func);
+        }
+        LLVMStripModuleDebugInfo(module);
+
```
rsdy: Some vertical whitespace and comments would be helpful here to understand what this piece of code does.

cargo-bpf/src/llvm.rs
```diff
@@ -0,0 +1,121 @@
...
+use llvm_sys::initialization::*;
+use llvm_sys::ir_reader::LLVMParseIRInContext;
+use llvm_sys::prelude::*;
+use llvm_sys::target::*;
+use llvm_sys::{LLVMAttributeFunctionIndex, LLVMAttributeReturnIndex, LLVMInlineAsmDialect::*};
+use std::ffi::{CStr, CString};
+use std::os::raw::c_char;
+use std::ptr;
+
+pub fn process_ir(input: &Path, output: &Path) -> Result<(), String> {
```
rsdy: It might make sense to break up this function to logical parts, or at least put in comments. It's not in a performance critical path, so I'm sure we could get away with simply splitting it up

alessandrod: Done

cargo-bpf/src/build.rs
alessandrod: I just copy/pasted since hopefully in the next few days that code is going to be deleted entirely once I figure out the passes to configure in llvm to compile directly from cargo-bpf without having to call out to opt/llc.

If you still want me to rework it, I'm happy to do that.

rsdy: I'm happy with that approach.

PR #53, opened 2020-03-04
Enable Sccache in CI

rsdy:
We should probably roll our own build containers out of band for bigger speedups.


alessandrod:
On ubuntu I think we can install sccache with snap, see https://snapcraft.io/install/sccache/ubuntu

PR #54, opened 2020-03-05
Fixed Typos

rtkaratekid:
"cargo bpf add" was providing an example that didn't compile.

Three simple changes fixed it.
kprobes -> kprobe
id -> pid
ctx -> regs.ctx

The default program added by "cargo bpf add" is now able to be compiled.

### DIFF
--- a/cargo-bpf/src/new_program.rs
+++ b/cargo-bpf/src/new_program.rs
@@ -111,7 +111,7 @@ program!(0xFFFFFFFE, "GPL");
 // #[map("syscall_events")]
 // static mut syscall_events: PerfMap<SomeEvent> = PerfMap::with_max_entries(1024);
 //
-// #[kprobe("syscall_enter")]
+// #[kprobe("do_sys_open")]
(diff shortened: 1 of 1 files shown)


alessandrod:
Thanks for the PR - looks good :) It looks like the entry point passed to `#[kprob]` is still wrong, it should be `#[kprobe("__x64_sys_clone")]` or any other syscall. Could you fix that too?

rtkaratekid:
@alessandrod sure thing, you mean like ```#[kprobe("__x64_do_sys_open")]```?
The current example is ```#[kprobe("sys_enter")]``` which is actually a tracepoint.

alessandrod:
See this https://github.com/redsift/ingraind/commit/1af61fe2109f5c3f6addc705c5e10a9abb464963

The prefix is `__<arch>_sys_` on newer kernels (>= 4.17) and `sys_` on older ones. Assuming that the latest popular distros have kernels newer than 4.17, we should probably go for `__x64_sys_open`

rsdy:
Thanks!

PR #66, opened 2020-07-23
Upgrade to LLVM 10

aquarhead:
(Forgot to mention the reasoning, [Rust 1.45 is using LLVM 10 now](https://github.com/rust-lang/rust/commit/82911b3bba76e73afe2881b732fe6b0edb35d5d3) which is the merge commit from [this PR](https://github.com/rust-lang/rust/pull/67759))

The code change should be trivial.

I think we might want to clean up the README a bit, for Ubuntu dependencies:

- Don't think we need `curl` `wget` or `gnupg2` at all?
- Don't think we need `ca-certificates-java` either?
- It seems `libelf` is enough (I just realized I didn't have `libelf-dev` - though installing it doesn't solve the load issue I have, and it certainly didn't affect `cargo bpf build`)
- `zlib1g` is needed (forgot about the error message but it's the first error I hit when running `cargo bpf build`)
- `linux-headers` need to pick an explicit version, otherwise `apt` (`apt-get` too) complains `E: Package 'linux-headers' has no installation candidate`

It might also be related to Ubuntu releases, I'm running on 20.04 (Desktop).

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ project](https://github.com/redsift/ingraind/tree/v1.0).
 In order to build some of the code here, you will need the following:
 
  * Linux 4.19+, with a build tree. The build tree is picked up from standard locations, or the `KERNEL_SOURCE` environment variable.
- * LLVM 9, or an LLVM version compatible with the Rust release you're using to build
+ * LLVM 10
  * The latest stable Rust compiler. We only promise to build with the latest stable and nightly compilers.
 
 # Getting started
@@ -70,7 +70,6 @@ sure you sync the git submodules necessary to build redbpf:
 
 Then install the dependencies for your distro before running the usual ritual.
 
-	cargo build --release
     cargo install --path cargo-bpf
 
 ## Ubuntu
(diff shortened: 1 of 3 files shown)


alessandrod:
Thanks! I'll take a look soon, I'll rebase my related changes on top of this.

rsdy:
This is great, thank you!

I'm wondering if it's worth making this change backwards compatible, or should we enforce a hard dependency on LLVM10. It looks like Bionic and newer Ubuntu versions have a proper `llvm-10` in the repo, and Arch already packages LLVM9 as `llvm-9`, so if we adjust the CI, there should be little impact in moving forward, and we don't want to support older Fedora anyway.

alessandrod:
I've tested this PR together with https://github.com/redsift/redbpf/pull/67 and things do work again with stable rust.

Keeping backwards compatibility isn't worth it since we operate on LLVM bitcode which isn't stable, so for things to work we must use the same LLVM version used by rustc which at the moment is 10.

The only thing that's missing to merge this is changing the CI images to install llvm-10 instead of llvm-9.

aquarhead:
Can confirm adding #67 works :)

> changing the CI images to install llvm-10 instead of llvm-9

While you're at it, can you also take a look at other dependencies? I'll follow up in this PR to further cleanup the dependencies list if you let me know the minimum set required.

rsdy:
We have some CI issues at the moment I'll need to take a look at. Thanks both for the contribution, I'll merge #67 along with this.

PR #68, opened 2020-07-27
cargo-bpf: load: don't unload programs too early

alessandrod:
This PR fixes a bug where the loader would accidentally get dropped early for programs that don't submit any events, causing programs to get unloaded.

Example program that used to trigger the bug:

```rust
#![no_std]
#![no_main]
use redbpf_probes::xdp::prelude::*;

program!(0xFFFFFFFE, "GPL");

#[xdp("block_port")]
pub fn block_port(ctx: XdpContext) -> XdpResult {
    match ctx.transport()? {
        t @ Transport::TCP(_) if t.dest() == 1111 => Ok(XdpAction::Drop),
        _ => return Ok(XdpAction::Pass),
    }
```

### DIFF
--- a/cargo-bpf/src/load.rs
+++ b/cargo-bpf/src/load.rs
@@ -7,7 +7,7 @@
 
 use crate::CommandError;
 
-use futures::stream::StreamExt;
+use futures::{future, stream::StreamExt};
 use hexdump::hexdump;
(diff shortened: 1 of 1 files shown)


aquarhead:
Can confirm this works for programs without maps :)

PR #70, opened 2020-08-07
[Fix #69] impl for multiple XDP attach

### LINKED ISSUE #69: Incorrect Drop after multiple `attach_xdp` calls
This line https://github.com/redsift/redbpf/blob/8c1d771bcd29b6e0415544ac4aaa0a3d5006519f/redbpf/src/lib.rs#L453 causes `Drop` to detach the XDP program from only the last interface.

I'm thinking to change the interface tracking to a Vec, but I'm not entirely sure the correct semantic for `attach_xdp`. The other option would be to detach from the previous interface so it only attaches once, though that doesn't match my use case. I'd be happy to attempt a PR if you have a preference.

This is discovered when I try to attach a program to all interfaces like this:

```rust
let interfaces = std::fs::read_dir("/sys/class/net").expect("failed to list interfaces");
for interface in interfaces {
  if let Ok(iface) = interface {
    if let Ok(ifa) = iface.file_name().into_string() {
      prog.attach_xdp(&ifa, xdp::Flags::default()).expect("failed to attach XDP program");
    }
  }
}
```

aquarhead:


rsdy:
This is great, thank you!

Issue #74, opened 2020-08-24
Build fails with kernel 5.7.0 on Debian


andrejtokarcik:
After #72 got merged I wanted to test the support for newer kernels with the version 5.7.0-2-amd64 available in Debian testing repositories. However, I encountered the following error:

```
tc@debian:~/redbpf$ cargo build
   [...]
   Compiling bpf-sys v1.1.0 (/home/tc/redbpf/bpf-sys)
   Compiling redbpf v1.1.0 (/home/tc/redbpf/redbpf)
   Compiling cargo-bpf v1.1.0 (/home/tc/redbpf/cargo-bpf)
   Compiling redbpf-probes v1.1.0 (/home/tc/redbpf/redbpf-probes)
   Compiling probes v0.1.0 (/home/tc/redbpf/redbpf-tools/probes)
   Compiling redbpf-tools v0.1.0 (/home/tc/redbpf/redbpf-tools)
error: failed to run custom build command for `redbpf-tools v0.1.0 (/home/tc/redbpf/redbpf-tools)`

Caused by:
  process didn't exit successfully: `/home/tc/redbpf/target/debug/build/redbpf-tools-105a7201ed085894/build-script-build` (exit code: 101)

--- stderr
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', bpf-sys/src/headers.rs:90:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

(I am doing `cargo build` from within a Git clone but I believe `cargo install cargo-bpf` would result in the same.)

The problem is that the Makefile used in `build_kernel_version()` (/lib/modules/5.7.0-2-amd64/build/Makefile) happens to contain just an `include` of the Makefile from the source headers tree, whereas the parsing procedure of `build_kernel_version()` expects the kernel version variables to be defined immediately in the given Makefile.

I see two possible solutions:

1. Use the source headers' Makefile instead, i.e. refer to the `source` field of the `KernelHeaders` struct when assembling the Makefile path, not the `build` field.
2. Run the `make` command to pre-parse the Makefile and resolve any indirections, as in andrejtokarcik@8500758ad94d679c7d2077caeb045245f0be15a0.

rsdy:
This is great, thanks. I'll try your branch on a few other distros and in the CI builders to see if there are any hidden dependencies this introduces. I'd very much like to merge your patch if it looks fine everywhere.

andrejtokarcik:
Sure, you're welcome!

[...]

PR #75, opened 2020-08-25
bpf-sys: Make build_kernel_version() more robust

andrejtokarcik:
Do not assume the kernel version variables to be defined directly
in the Makefile from the build headers tree.  Instead, let `make`
parse the Makefile, resolve all potential includes of other
Makefiles and print the variable definitions.


rsdy:
Awesome, thanks. Resolves #74

PR #90, opened 2020-10-02
Add StackTrace support

quininer:
Hi, I'm trying to [use redbpf to track memory allocation](https://github.com/quininer/redbpf-memtrace), support stack map would be very helpful.

### DIFF
--- a/redbpf-probes/src/maps.rs
+++ b/redbpf-probes/src/maps.rs
@@ -237,7 +237,7 @@ impl StackTrace {
         }
     }
 
-    pub unsafe fn stackid(&mut self, ctx: *mut pt_regs, flag: u64) -> Result<c_int, c_int> {
+    pub unsafe fn stack_id(&mut self, ctx: *mut pt_regs, flag: u64) -> Result<c_int, c_int> {
         let ret = bpf_get_stackid(ctx as _, &mut self.def as *mut _ as _, flag);
         if ret >= 0 {
             Ok(ret)
--- a/redbpf-probes/src/maps.rs
+++ b/redbpf-probes/src/maps.rs
@@ -210,3 +210,39 @@ impl<T> PerfMap<T> {
         );
     }
 }
+
+// TODO Use PERF_MAX_STACK_DEPTH
+const BPF_MAX_STACK_DEPTH: usize = 127;
+
+#[repr(transparent)]
+pub struct StackTrace {
+    def: bpf_map_def
+}
+
+#[repr(C)]
+struct BpfStackFrames {
+    ip: [u64; BPF_MAX_STACK_DEPTH]
+}
+
+impl StackTrace {
+    pub const fn with_capacity(cap: u32) -> Self {
+        StackTrace {
+            def: bpf_map_def {
+                type_: bpf_map_type_BPF_MAP_TYPE_STACK_TRACE,
+                key_size: mem::size_of::<u32>() as u32,
+                value_size: mem::size_of::<BpfStackFrames>() as u32,
+                max_entries: cap,
+                map_flags: 0
+            }
+        }
+    }
+
+    pub unsafe fn stackid(&mut self, ctx: *mut pt_regs, flag: u64) -> Result<c_int, c_int> {
+        let ret = bpf_get_stackid(ctx as _, &mut self.def as *mut _ as _, flag);
+        if ret >= 0 {
+            Ok(ret)
+        } else {
+            Err(ret)
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


rsdy:
This is really cool! Thanks! I'll have some time to take a proper look during the weekend.

redbpf/src/lib.rs
```diff
@@ -820,6 +832,43 @@ impl<K: Clone, V: Clone> Iterator for MapIter<'_, '_, K, V> {
...
+
+            if ret == 0 {
+                Some(value.assume_init())
+            } else {
+                None
+            }
+        }
+    }
+
+    pub fn remove(&mut self, id: libc::c_int) -> bool {
```
rsdy, line 860: `HashMap` has a `delete()` that is consistent with the bpf terminology, so I'd prefer using that convention instead of `remove`.

redbpf/src/lib.rs
```diff
@@ -820,6 +832,43 @@ impl<K: Clone, V: Clone> Iterator for MapIter<'_, '_, K, V> {
...
+        }
+    }
+
+    pub fn get(&mut self, id: libc::c_int) -> Option<BpfStackFrames> {
+        unsafe {
+            let mut value = MaybeUninit::uninit();
+
+            let ret = bpf_sys::bpf_lookup_elem(
+                self.base.fd,
+                &id as *const libc::c_int as *mut libc::c_int as _,
```
rsdy, line 848: An easier way to do this might be at https://github.com/redsift/redbpf/blob/78f6ac9113d9cfad44e4b4dccc06e1fba763332e/redbpf/src/lib.rs#L814 .

redbpf/src/lib.rs
```diff
@@ -820,6 +832,43 @@ impl<K: Clone, V: Clone> Iterator for MapIter<'_, '_, K, V> {
...
+    }
+
+    pub fn remove(&mut self, id: libc::c_int) -> bool {
+        unsafe {
+            let ret = bpf_sys::bpf_delete_elem(
+                self.base.fd,
+                &id as *const libc::c_int as *mut libc::c_int as _,
+            );
+
+            ret == 0
```
rsdy, line 867: This is great! Can you please modify `HashMap::delete` to be consistent with reporting the error from the kernel? I think the best would probably be to return `Result<()>`, and `Err(Error::Map)` for failure in both cases.

redbpf-probes/src/maps.rs
```diff
@@ -210,3 +210,39 @@ impl<T> PerfMap<T> {
...
+    def: bpf_map_def
+}
+
+#[repr(C)]
+struct BpfStackFrames {
+    ip: [u64; BPF_MAX_STACK_DEPTH]
+}
+
+impl StackTrace {
+    pub const fn with_capacity(cap: u32) -> Self {
```
rsdy, line 228: Can we call this `with_max_entries` to stay consistent?

rsdy (changes requested):
Apart from naming and the function signature stuff, it looks good. Happy to merge it once you make these changes!

redbpf/src/lib.rs
quininer: Change `HashMap::delete` will cause API break, maybe we can add a `HashMap::delete2`?

redbpf-probes/src/maps.rs
```diff
@@ -210,3 +210,39 @@ impl<T> PerfMap<T> {
...
+                type_: bpf_map_type_BPF_MAP_TYPE_STACK_TRACE,
+                key_size: mem::size_of::<u32>() as u32,
+                value_size: mem::size_of::<BpfStackFrames>() as u32,
+                max_entries: cap,
+                map_flags: 0
+            }
+        }
+    }
+
+    pub unsafe fn stackid(&mut self, ctx: *mut pt_regs, flag: u64) -> Result<c_int, c_int> {
```
alessandrod, line 240: I think this should be named stack_id

alessandrod (commented):
stackid vs stack_id thing aside, this looks good. I'm planning to rework redbpf::Map a bit so it would be good to merge this before I do that.

redbpf/src/lib.rs
alessandrod: I think it's not worth adding a separate method nor breaking API just for this. I'm planning to overhaul error handling which is now a bit messy soon, so I can fix this when I get to that.

rsdy (approved):

rsdy:
Thanks both, I'm merging it in this state.

PR #92, opened 2020-10-15
Add examples

rhdxmr:
redbpf-examples is added for showing simple BPF programs which are not enough quality to be redbpf-tools. And it is expected that newbies can get to know what they can do with redbpf by studying code of redbpf-examples.

I am a newbie of redbpf and I hope this example code are helpful to other newbies.
And I anticipate that other users of redbpf can put their bpf programs into this example directory easily.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,2 @@
 [workspace]
-members = ["bpf-sys", "redbpf", "redbpf-probes", "redbpf-macros", "cargo-bpf", "redbpf-tools"]
+members = ["bpf-sys", "redbpf", "redbpf-probes", "redbpf-macros", "cargo-bpf", "redbpf-tools", "examples/example-probes", "examples/example-userspace"]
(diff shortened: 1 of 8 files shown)


rsdy:
I think I would prefer if this was in an `examples` folder, or simply part of `redbpf-tools`. Using `examples` gives us the benefit that cargo [is able to handle it straight from the command line](https://doc.rust-lang.org/cargo/commands/cargo-run.html#examples).

rsdy (changes requested):
I think I would prefer if this was in an `examples` folder, or simply part of `redbpf-tools`. Using `examples` gives us the benefit that cargo [is able to handle it straight from the command line](https://doc.rust-lang.org/cargo/commands/cargo-run.html#examples).

rhdxmr:
@rsdy OK I'll modify the PR to use `examples` Thanks,

rsdy (approved):

rhdxmr:
hello @rsdy 
I changed redbpf-examples as examples and it is now available with cargo run --example
Can you review this?
Thanks,

PR #96, opened 2020-10-19
Implement post processing and code generation with libLLVM

alessandrod:
This PR makes cargo-bpf use libLLVM directly to generate code instead of using opt + llc. This is more robust and gives us the possibility to implement more custom optimizations in the future. Also, now upgrading LLVM version is matter of updating the `llvm-sys` version in Cargo.toml.

Note that by using libLLVM we use the legacy pass manager (the only one available), instead of the new pass manager used by opt and llc, so the compilation pipeline isn't 100% the same. I have tested with ingraind, redbpf-tools and snuffy and everything seems to work ok.

The PR is based on https://github.com/redsift/redbpf/pull/95 so that should be merged first.

@rsdy it would be great if you could run this on the ingraind CI and see how things look in the terraform jobs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,8 @@ project](https://github.com/redsift/ingraind/tree/v1.0.0).
 In order to build some of the code here, you will need the following:
 
  * Linux 4.19+, with a build tree. The build tree is picked up from standard locations, or the `KERNEL_SOURCE` environment variable.
- * LLVM 10
+ * LLVM 11. For debian based distros that don't have LLVM 11, you can find packages at https://apt.llvm.org. For rpm based ones, you can get it from Fedora Rawhide.
  * The latest stable Rust compiler. We only promise to build with the latest stable and nightly compilers.
-
-NOTE: the latest `nightly` compilers will use LLVM 11, which doesn't ship with most distributions. Use a stable compiler for the time being.
 
 # Getting started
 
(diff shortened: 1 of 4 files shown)


rsdy:
It might take a few days for me to actually spin up an ARM machine, as the Terraform testbench is temporarily inactive for now. Will ping you back to see if there are any errors.

rsdy:
One considerable difficulty testing this is that Rust is now built on LLVM11, but no distribution actually ships that, including Arch, so this will inevitably fail everywhere.

rsdy:
Looks like it works on Arch with the Rust compiler coming from Pacman. We need to see what we can do with other distributions, but realistically, I'm not convinced that many people will use distribution packages over `rustup`.

alessandrod:
> Looks like it works on Arch with the Rust compiler coming from Pacman. We need to see what we can do with other distributions, but realistically, I'm not convinced that many people will use distribution packages over `rustup`.

Ok but just to clarify, deployment wise this PR changes nothing. We require LLVM 11 regardless of whether we use llc-11 or link to libLLVM.

There's https://apt.llvm.org for debian based distributions, was there something similar for rpm based distros?

Also, what kernel version did  you try with?

rsdy:
We can pull in llvm11 from Fedora Rawhide

rsdy:
@alessandrod if you're ready with this I've adjusted the CI to build with LLVM11. We're seeing a bunch of [failures](https://app.circleci.com/pipelines/github/redsift/redbpf/272/workflows/a53de0ca-5446-4f30-b1ae-8e8850072181/jobs/553), though, so I think this patchset will be required for proper support.

alessandrod:
Yeah I have one extra change I'm going to push on top of this. Will push it later today!

alessandrod:
Alright I think this is ready to be merged now

rsdy (approved):

Issue #56, opened 2020-03-19
Difference in section name interpretation, compared to libbpf


tilpner:
Hi, I modified redbpf to load `BPF_PROG_TYPE_CGROUP_SKB` programs, but there's a design decision to be made before sending a PR.

libbpf [has a mapping of section names to program parameters](https://github.com/redsift/libbpf/blob/4d86cae4f09f3db67d6d0543f94546f85f7b2c38/src/libbpf.c#L4518), and I think it would be good to keep redbpfs interpretation as close as possible, to prevent confusion by e.g. slightly different uses of the second part.

Should redbpf try to mirror the libbpf list, restricted to implemented types, of course?

Specifically, with a section name of `cgroup_skb/egress`, libbpf will match it to [a program type of `BPF_PROG_TYPE_CGROUP_SKB`, and an expected attachment type of `BPF_CGROUP_INET_EGRESS`](https://github.com/redsift/libbpf/blob/4d86cae4f09f3db67d6d0543f94546f85f7b2c38/src/libbpf.c#L4534-L4535), but redbpf will [use the `egress` part as a name](https://github.com/redsift/redbpf/blob/a8cbba0b2b0bf892068ac4c521f51f43fb2c0874/redbpf/src/lib.rs#L420), and thus can't select an expected attachment type.
libbpf instead seems to [use the symbol name as the program name](https://github.com/redsift/libbpf/blob/4d86cae4f09f3db67d6d0543f94546f85f7b2c38/src/libbpf.c#L466-L48).

Assigning meaning to the second part might break some existing code, but does reduce future surprises for new users. Consistency with libbpf helps when following documentation meant for libbpf.

alessandrod:
> Hi, I modified redbpf to load `BPF_PROG_TYPE_CGROUP_SKB` programs, but there's a design decision to be made before sending a PR.

Awesome! Please keep in mind that in the process of adding support for uprobes, I'm currently refactoring `redbpf::Program` and friends and I'm aiming to send a PR sometime in the next 24hrs, then get it merged early next week. If you're almost ready to send a PR, we probably want to try to get it merged before we merge mine as my changes are pretty invasive (I'm happy to rebase on your work).

[...]

PR #97, opened 2020-10-19
Add support for `tc` actions

aquarhead:
This work has been on my fork a couple month now, and I had a conversation with @alessandrod about this when he said he would like to add support for `tc` programs, that is also almost a month ago now... Forgive my laziness but anyway here I am ready to kick off this effort.

You can see in this PR that I've moved the `SkBuff` out of the `socket_filter` module, since `tc` programs use it for input as well, and (I believe) I followed existing example for other shared data structures for probes. I also plan to add a few more utility functions in `SkBuff` for example to test for specific protocol, and jump to packet data start - since IP and TCP header has some peculiar header length calculation that I spent quite a while to figure out. 

I picked the macro name `tc_action` because as I understand, there're also "classifiers" for `tc`, although I haven't really figure it out what they do or whether they're really different than "action" programs. As far as I remember, they have different return codes, so I believe it makes sense to separate them based on that.

Note with only this PR, the output ELF still cannot be loaded by `tc`. At least two problems that I've found workarounds:

- Strip `.text` section
- Rename `maps` section

As a complete working example, you can see [this project](https://github.com/aquarhead/protect-the-rabbit), and the workarounds are codified [here](https://github.com/aquarhead/protect-the-rabbit/blob/master/Makefile.toml#L10-L18).

I think it's possible that we integrate these extra LLVM tools, but I'd like to see your opinion first. Also I'd really like to understand what's in the `.text` section anyway...

Another potential issue is how to use multiple maps if `tc` only expects a single section, though it might not be an issue but I don't think I've tried it.

On a sidenote, I submitted a lightning talk on exactly this work for the eBPF summit, and it got accepted, I'd really appreciate it if you would like to see my draft (I need to prepare a recorded video before Friday).

### DIFF
--- a/redbpf-macros/src/lib.rs
+++ b/redbpf-macros/src/lib.rs
@@ -363,3 +363,29 @@ pub fn socket_filter(attrs: TokenStream, item: TokenStream) -> TokenStream {
 
     probe_impl("socketfilter", attrs, wrapper, name).into()
 }
+
+/// Define [tc action BPF programs](https://man7.org/linux/man-pages/man8/tc-bpf.8.html)
+#[proc_macro_attribute]
+pub fn tc_action(attrs: TokenStream, item: TokenStream) -> TokenStream {
+    let item = parse_macro_input!(item as ItemFn);
+    let name = item.sig.ident.to_string();
+    let ident = item.sig.ident.clone();
+    let outer_ident = Ident::new(&format!("outer_{}", ident), Span::call_site());
+    let wrapper = parse_quote! {
+        fn #outer_ident(skb: *const ::redbpf_probes::bindings::__sk_buff) -> i32 {
+            let skb = ::redbpf_probes::socket::SkBuff { skb };
+            return match #ident(skb) {
+                Ok(::redbpf_probes::tc::TcAction::Ok) => 0,
+                Ok(::redbpf_probes::tc::TcAction::Shot) => 2,
+                Ok(::redbpf_probes::tc::TcAction::Unspec) => -1,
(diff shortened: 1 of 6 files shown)


rsdy:
Hey, it's great to see this, thanks!
I'd be happy to review the talk draft, if you message me here or on Twitter, my DMs are open.

rsdy (approved):
Really good quality PR, I have no objections to merging this as is, if you're happy with it. Thanks a lot!

alessandrod:
I think both issues can be fixed without incompatible changes to the current toolchain.

The `.text` section has to do with https://github.com/alessandrod/redbpf/blob/685c51862f58d89b399958e1cb126fa325fb4059/cargo-bpf/src/llvm.rs#L182. Since we have no way to provide the library info, and since we ask rustc to do the first step of the compilation for the host target triple, llvm ends up thinking that the target provides bcmp when it doesn't. So to make things work we inject our own bcmp (and other builtins), that end up in .text. I'm working on a workaround for this right now.

The map thing we just need to decide the API. Currently `#[map("foo")]` outputs an ELF section named "maps/foo". we could have `#[map(name="")]` or something to output just "maps".

aquarhead:
Just remembered, I believe `tc` also supports attaching already loaded programs - which means it doesn't need to interpret the ELF file by itself - so maybe we can also just provide/extend our loader to load the program and maps and don't necessarily need to "fix" the ELF file?

rsdy:
Let's move this loader discussion to #56 . I'm merging this. Thanks both!

PR #107, opened 2020-11-13
allows specifying the link_section for a map

### LINKED ISSUE #105: tc BPF programs only understand "maps" section
BPF programs loaded by `tc` only understand  a single section called `maps`. The current [`map` macro](https://ingraind.org/api/redbpf_macros/attr.map.html) which places each map in its own `maps/<name>` section conflicts with how `tc` is used. To fix this one must manually rename the section with something like `llvm-objcopy --rename-section`, however this starts to fall over and become tedious when more than one map is required (you end up with multiple sections simply named `maps` which again confuses `tc`).

I'd be happy to submit a PR to either add a `tc_map` macro which places all defined maps into the same `maps` section, or alternatively add logic to the current macro and that if a map name is omitted (i.e. `#[map]`) its placed in the global `maps` section. I'd just need to know which route you would like to go, and if there are any other concerns to know about.

kbknapp:
This commit changes the `map` attribute macro to allow specifying a section
name manually, or using the item name when explict names are omitted.

For example, the *old* version:

```rust
static mut counts: PerfMap<u64> = PerfMap::with_max_entries(10240);
```

Would place the map in the `maps/counts` section of the resulting ELF
binary. Note the repetition of the names.

The new style infers the name from the item:

```rust
static mut counts: PerfMap<u64> = PerfMap::with_max_entries(10240);
```

Which still places the map in the `maps/counts` section of the binary.

However, in some cases the default naming scheme does not work (such as
in `tc` BPF programs). If the link section must be supplied manually,
one can use this method:

```rust
static mut counts: PerfMap<u64> = PerfMap::with_max_entries(10240);
```

In this case the map will be placed in the explicit `foo` section of the
binary.

Closes #105

### DIFF
--- a/redbpf-macros/Cargo.toml
+++ b/redbpf-macros/Cargo.toml
@@ -17,6 +17,9 @@ proc-macro2 = "1.0"
 syn = {version = "1.0", features = ["full"] }
 quote = "1.0"
 
+[build-dependencies]
+rustc_version = "0.3.0"
+
 [dev-dependencies]
 # redbpf-probes is needed by doctests
 redbpf-probes = { version = "^1.2.0", path = "../redbpf-probes" }
--- a/redbpf-macros/build.rs
+++ b/redbpf-macros/build.rs
@@ -1,0 +1,20 @@
+extern crate rustc_version;
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    // Set cfg flags depending on release channel
+    match version_meta().unwrap().channel {
+        Channel::Stable => {
+            println!("cargo:rustc-cfg=RUSTC_IS_STABLE");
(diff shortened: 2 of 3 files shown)


alessandrod (changes requested):
This looks ok but it breaks the current API so it can't be merged until we release 2.x. 

In order to merge it now it needs to be changed to still support the old `#[map("literal")]` form, and we should somehow output a deprecation warning when the old form is used.

kbknapp:
Sounds good, I can add that later today.

rsdy (approved):

rsdy:
I'm super happy with this. Really good quality code! Thanks!

PR #111, opened 2020-11-25
Reduces bindgen child-deps + fixes clippy lints

kbknapp:
This PR is in two commits:

- Reducing compile time deps of `bindgen` by removing default features (these deps are not required by anything `redbpf` is doing)
- Fixes a bunch of clippy lints.

The commits have decent messages to explain the details.

Tests are currently failing on `master` so I wasn't able to check that, but all tests that pass on master still pass here. I also compiled against current stable and current nightly.

### DIFF
--- a/bpf-sys/Cargo.toml
+++ b/bpf-sys/Cargo.toml
@@ -16,5 +16,5 @@ libc = "0.2"
 
 [build-dependencies]
 cc = "1.0"
-bindgen = "0.55"
+bindgen = {version = "0.55", default-features = false, features = ["runtime"]}
 libc = "0.2"
--- a/bpf-sys/build.rs
+++ b/bpf-sys/build.rs
@@ -9,7 +9,7 @@
 use std::env;
 use std::path::PathBuf;
 
-const KERNEL_HEADERS: [&'static str; 6] = [
+const KERNEL_HEADERS: [&str; 6] = [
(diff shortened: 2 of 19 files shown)


rsdy:
This looks great, thanks! It breaks a few public APIs, so we'll definitely need a new minor version when we merge it, and most likely will break ingraind too. Let me run a few tests as well before merging!

kbknapp:
> It breaks a few public APIs

I think the only spot the public API changes is in `bpf-sys` where a return goes from `Result<T, ()>` to `Option<T>`. I think it'd be fine to roll that change back until an eventual 2.0 as its just a pedantic change, there is nothing inherently *wrong* with `Result<T, ()>`, its just less idiomatic.

Let me know if you'd like me to roll that back :wink:

rsdy:
@kbknapp That might be the best approach. I agree that the API is not idiomatic, although to be fair using `MaybeUninit` in specifically `uname()` would potentially give us a nicer implementation. If you roll back those changes, I can merge this quickly.

kbknapp:
Sorry for the delay; with the holidays it took me a minute to sit down at a computer. I've rolled back those changes, but please do a double check and make sure I didn't miss any. Let me know if any additional changes are required.

rsdy:
This looks great! Thanks a lot!

rsdy (approved):

PR #115, opened 2020-12-03
Skips re-run of build.rs if C source doesnt change

kbknapp:
This commit tells `cargo` not to run the `build.rs` scripts in the
following crates if the specified C source files have not changed:

- `redbpf-probes`
- `bpf-sys`
- `redbpf-tools`
- `redbpf-tools/probes`

Prior to this commit `cargo` would re-run the build scripts (which are
generating language bindings to C libraries) on each compile/check. This
drastically slows down the dev cycle, especially if editors want to run
checks after each change/save.

On my desktop Threadripper 2950 (16 cores) with an NVMe drive, the time
spent generating bindings on *every* build was roughly 91 seconds in
debug mode in the worst case. But even cutting out the outlier 
(`redbpf-tools(/probes)`), each crate would be in the 8 second range which 
is still high for certain editors which re-run checks constantly.

This change cuts that wait time out entirely unless the C source files
change.

### DIFF
--- a/bpf-sys/build.rs
+++ b/bpf-sys/build.rs
@@ -26,8 +26,23 @@ pub mod headers {
     include!(concat!(env!("CARGO_MANIFEST_DIR"), "/src/headers.rs"));
 }
 
+fn rerun_if_changed_dir(dir: &str) {
+    println!("cargo:rerun-if-changed={}/", dir);
+    for ext in &["c", "h", "bash", "map", "md", "rst", "sh", "template"] {
+        glob::glob(&format!("./{}/**/*.{}", dir, ext))
+            .expect("Failed to glob for source files from build.rs")
+            .filter_map(|e| e.ok())
+            .for_each(|path| println!("cargo:rerun-if-changed={}", path.to_string_lossy()));
+    }
+}
+
 fn main() {
     println!("cargo:rustc-link-lib=static=bpf");
+    for dir in &["bcc", "libbpf", "libelf"] {
+        rerun_if_changed_dir(dir);
+    }
+    println!("cargo:rerun-if-changed=bpfsys-musl.h");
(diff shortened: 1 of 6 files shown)


bpf-sys/build.rs
```diff
@@ -28,6 +28,11 @@ pub mod headers {
 
 fn main() {
     println!("cargo:rustc-link-lib=static=bpf");
+    println!("cargo:rerun-if-changed=bcc/");
+    println!("cargo:rerun-if-changed=libbpf/");
+    println!("cargo:rerun-if-changed=libelf/");
+    println!("cargo:rerun-if-changed=bpfsys-musl.h");
+    println!("cargo:rerun-if-changed=libbpf_xdp.h");
```
alessandrod, line 36: I don't think `rerun-if-changed=directory/` does the right thing, see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath

redbpf-tools/build.rs
```diff
@@ -4,6 +4,8 @@ use std::path::{Path, PathBuf};
 use cargo_bpf_lib as cargo_bpf;
 
 fn main() {
+    println!("cargo:rerun-if-changed=probes/");
```
alessandrod, line 7: This is unnecessary, at the bottom we're already printing a rerun-if-changed for each probe file

alessandrod (changes requested):
The direction is good but I think that most file systems won't update the mtime of a directory when something inside it changes. I think we need to glob and print one line for each file.

kbknapp:
Great points, I'll make the changes later today once I get to a computer.

kbknapp:
Changes have been made. Let me know if you see something else that should change :wink:

alessandrod (approved):
Thank you! Looks good now.

PR #116, opened 2020-12-05
Add support for .data .rodata and .bss

alessandrod:
Load .data .rodata and .bss as ARRAY maps of one item: the section data. Then during relocation patch instructions to point inside the map.

No more `let key = 1; f(&key);`

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -139,6 +139,7 @@ pub struct Map {
     pub kind: u32,
     fd: RawFd,
     config: bpf_map_def,
+    section_data: bool,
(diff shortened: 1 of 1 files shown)


kbknapp:
I had some open questions in URLO and the Rust Zulip about this exact issue, so I'll be very excited to see this merge!

alessandrod:
> I had some open questions in URLO and the Rust Zulip about this exact issue, so I'll be very excited to see this merge!

It's new and just getting started but now there's also a redbpf room on matrix https://app.element.io/#/room/#redbpf:rustch.at

PR #121, opened 2020-12-17
Make kernel sources dir configurable

vlad9486:
This close an [issue](https://github.com/redsift/redbpf/issues/119).

### DIFF
--- a/bpf-sys/src/headers.rs
+++ b/bpf-sys/src/headers.rs
@@ -61,8 +61,16 @@ pub fn running_kernel_version() -> Option<String> {
     })
 }
 
-pub fn build_kernel_version() -> Result<KernelVersion, Box<dyn Error>> {
-    let KernelHeaders { source: _, build } = kernel_headers_path()?;
+pub fn build_kernel_version(source_dir: Option<&Path>) -> Result<KernelVersion, Box<dyn Error>> {
+    let KernelHeaders { source: _, build } = if let Some(source_dir) = source_dir {
+        KernelHeaders {
+            source: source_dir.to_owned(),
+            build: source_dir.to_owned(),
+        }
+    } else {
+        kernel_headers_path()?
+    };
+
     let make_db = Command::new("make")
                           .arg("-qp")
                           .arg("-f")
--- a/cargo-bpf/src/build.rs
+++ b/cargo-bpf/src/build.rs
@@ -71,7 +71,13 @@ impl From<Error> for CommandError {
     }
 }
 
-fn build_probe(cargo: &Path, package: &Path, target_dir: &Path, probe: &str) -> Result<(), Error> {
+fn build_probe(
+    cargo: &Path,
+    package: &Path,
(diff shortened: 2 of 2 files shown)


bpf-sys/src/headers.rs
```diff
@@ -61,8 +61,16 @@ pub fn running_kernel_version() -> Option<String> {
     })
 }
 
-pub fn build_kernel_version() -> Result<KernelVersion, Box<dyn Error>> {
-    let KernelHeaders { source: _, build } = kernel_headers_path()?;
+pub fn build_kernel_version(source_dir: Option<&Path>) -> Result<KernelVersion, Box<dyn Error>> {
```
rsdy, line 64: I think it would be better if this version was called `build_kernel_version_from`, and the original `build_kernel_version` function provided a default wrapper. This means we can avoid breaking API, and the resulting code will also be cleaner.

cargo-bpf/src/build.rs
```diff
@@ -148,10 +155,20 @@ fn build_probe(cargo: &Path, package: &Path, target_dir: &Path, probe: &str) ->
...
 pub fn build(
+    cargo: &Path,
+    package: &Path,
+    target_dir: &Path,
+    probes: Vec<String>,
+) -> Result<(), Error> {
+    build_ext(cargo, package, target_dir, probes, None)
+}
+
+pub fn build_ext(
```
rsdy, line 166: I think this is better called  `build_with_kernel`

cargo-bpf/src/build.rs
```diff
@@ -148,10 +155,20 @@ fn build_probe(cargo: &Path, package: &Path, target_dir: &Path, probe: &str) ->
...
     cargo: &Path,
     package: &Path,
     target_dir: &Path,
     mut probes: Vec<String>,
+    kernel_source_dir: Option<&Path>,
```
rsdy, line 171: I'd prefer getting rid of propagating `Option<>` parameters in functions, and instead wrapping default behaviour with specific implementations at the top of the call tree.

cargo-bpf/src/build.rs
```diff
@@ -71,7 +71,13 @@ impl From<Error> for CommandError {
     }
 }
 
-fn build_probe(cargo: &Path, package: &Path, target_dir: &Path, probe: &str) -> Result<(), Error> {
+fn build_probe(
```
rsdy, line 74: I'd prefer no `Option` parameters to functions, and following the `build_probe_with_kernel` pattern for direct invocation, along with the wrapper `build_probe` that preserves the existing logic, same as above.

rsdy (changes requested):
Thanks for this. I think the feature is useful, but the coding style could be improved. I hope that my comments make sense, but please do ask if you need more specific examples of what I mean!

rsdy:
Closing as stale.
@vlad9486  if you have any interest in picking up this PR, please re-open.

PR #125, opened 2021-01-17
cargo-bpf debug info.

dvc94ch:
Emits btf type information required for relocating maps contained in the `.maps` section.


dvc94ch:
Would be nice, but also don't need it anymore

PR #128, opened 2021-03-06
Add array and percpu array maps

rhdxmr:
Add `Array` and `PerCpuArray` structures for BPF programs and userspace programs.
Add `biolatpcts` example program.


redbpf/src/cpus.rs
```diff
@@ -10,9 +10,20 @@ use std::io::Error;
 use std::str::FromStr;
 
 const SYS_CPU_ONLINE: &str = "/sys/devices/system/cpu/online";
+const SYS_CPU_POSSIBLE: &str = "/sys/devices/system/cpu/possible";
 
 pub type CpuId = i32;
 
+lazy_static! {
+    // possible cpu is static after kernel boot time. So it's OK to cache the
```
rsdy, line 18: What happens if you use some cgroup magic to set cpu affinity? Eg. libvirt can [do this](https://docs.fedoraproject.org/en-US/Fedora/13/html/Virtualization_Guide/ch25s06.html), meaning the host will not schedule tasks on the logical CPU, possibly affecting this too.

redbpf-probes/src/maps.rs
```diff
@@ -21,87 +21,190 @@ use cty::*;
...
+                        value as *const _ as *const c_void,
+                        BPF_ANY.into(),
+                    );
+                }
+            }
         }
-    }
+    };
 }
+define_hashmap!(
```
rsdy, line 178: These are awesome. Thanks!

rsdy (changes requested):
This is really high quality stuff, so I don't have any major complaints. However, I'm unsure about caching `/sys/` contents in userspace. Best case, it's a minor gain, worst case it will result in hard-to-debug issues.

redbpf/src/cpus.rs
rhdxmr: Oh that's a good point, that I've never considered before.
I agree with you. Performance gain would be little, but perhaps there exist many unknown corner cases the assumption becomes incorrect.
Before removing caching logic, I am going to investigate the situations you mentioned. Thanks,

rhdxmr:
Hello @rsdy 
I modified what you corrected and re-push the commits.
I feel sorry because I am too late. Phew, I've been too busy to progress this PR recently.

And I asked a question about /sys/devices/system/cpu/possible at kernel-janitors but it seems that I asked it to the wrong mailing list. Hmm.. (https://marc.info/?l=kernel-janitors&m=161598682220098&w=2)

rhdxmr:
I asked the question to kernelnewbies mailing list and I received an answer. haha

http://lists.kernelnewbies.org/pipermail/kernelnewbies/2021-March/021440.html

rsdy:
Oh this is great! Thanks for the changes. Also, super exhaustive and interesting reply from gkh!

rsdy (approved):

PR #131, opened 2021-03-27
Upgrade version of Tokio to ^1.0.1

### LINKED ISSUE #129: Newer version of Tokio
Any thoughts on updating the version of Tokio to 1.3? Importing 1.3 and 0.2 into my crate is not the end of the world, but having one Tokio library to work with would be nice.

rhdxmr:
This resolves #129 issue.
All example programs work as expected.
All packages of workspace are compiled successfully with stable rust 1.51.0.

### DIFF
--- a/cargo-bpf/Cargo.toml
+++ b/cargo-bpf/Cargo.toml
@@ -24,7 +24,7 @@ toml_edit = { version = "0.2", optional = true }
 bpf-sys = { version = "^1.3.0", path = "../bpf-sys", optional = true }
 redbpf = { version = "^1.3.0", path = "../redbpf", default-features = false, optional = true }
 futures = { version = "0.3", optional = true }
-tokio = { version = "^0.2.4", features = ["rt-core", "io-driver", "macros", "signal"], optional = true }
+tokio = { version = "^1.0.1", features = ["rt", "macros", "signal"], optional = true }
 hexdump = { version = "0.1", optional = true }
 libc = {version = "0.2.66", optional = true}
 llvm-sys = { version = "110", optional = true}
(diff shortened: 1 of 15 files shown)


rhdxmr:
And MapIo is marked as deprecated because it implements `mio::unix::Evented` which is not provided by newer version of mio.

rhdxmr:
@rsdy I found that the compiling failed under Fedora but succeeded under Ubuntu.
(My local machine uses Ubuntu 20.10 based PopOS)

It seems that `AsRawFd` trait for `RawFd` was removed in Apr 2017 by this PR
https://github.com/rust-lang/rust/pull/41035

And the trait was resurrected in Oct 2020 by this PR
https://github.com/rust-lang/rust/pull/76969

A new structure implementing `AsRawFd` is required to satisfy `AsyncFd<T: AsRawFd>`.
It's worth avoiding `AsyncFd<RawFd>` not to cause harm to existing RedBPF users who use rust < 1.48.

rsdy:
>  It's worth avoiding AsyncFd<RawFd> not to cause harm to existing RedBPF users who use rust < 1.48.

I don't think it's worth it, as it's quite an old release anyway. Realistically with Rust moving to newer LLVM versions, we're going to need to drop support for these anyway. I think targeting the latest stable compiler is reasonable.

rhdxmr:
@rsdy OK. That's more convenient way to develop this.
Then the build environment of fedora rawhide is needed to fixed to succeed the compilation?

rsdy (approved):
I think this is a good first step. Removing deprecations can come at a later stage.

rsdy:
Actually all the builds fail, and Fedora too. The fact that Ubuntu builds go through is a separate bug in the CircleCI files.

rhdxmr:
@rsdy Is there something that should be done in order to merge this PR?

Issue #130, opened 2021-03-22
Error building against musl


andrewrynhard:
I couldn't find a great example on how to compile against `musl`. Using archlinux (after installing `musl` and the build deps) I am seeing:

```
...
  TARGET = Some("x86_64-unknown-linux-musl")
  OPT_LEVEL = Some("z")
  HOST = Some("x86_64-unknown-linux-gnu")
  CC_x86_64-unknown-linux-musl = None
  CC_x86_64_unknown_linux_musl = None
  TARGET_CC = None
  CC = None
  CROSS_COMPILE = None
  CFLAGS_x86_64-unknown-linux-musl = None
  CFLAGS_x86_64_unknown_linux_musl = None
  TARGET_CFLAGS = None
  CFLAGS = None
  CRATE_CC_NO_DEFAULTS = None
  DEBUG = Some("false")
  CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")
...
  running: "musl-gcc" "-Os" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "libbpf/include/uapi" "-I" "libbpf/include" "-I" "bcc" "-I" "libelf" "-I" "." "-I" "/usr/lib/modules/5.11.8-arch1-1/build/arch/x86/include/generated/uapi" "-I" "/usr/lib/modules/5.11.8-arch1-1/build/arch/x86/include/uapi" "-I" "/usr/lib/modules/5.11.8-arch1-1/build/arch/x86/include/" "-I" "/usr/lib/modules/5.11.8-arch1-1/build/include/generated/uapi" "-I" "/usr/lib/modules/5.11.8-arch1-1/build/include/uapi" "-I" "/usr/lib/modules/5.11.8-arch1-1/build/include" "-Wall" "-Wextra" "-Wno-sign-compare" "-Wno-int-conversion" "-Wno-unused-parameter" "-Wno-unused-result" "-Wno-format-truncation" "-Wno-missing-field-initializers" "-include" "bpfsys-musl.h" "-include" "linux/stddef.h" "-DCOMPAT_NEED_REALLOCARRAY=1" "-o" "/usr/local/target/x86_64-unknown-linux-musl/release/build/bpf-sys-08ba1431d87e1df6/out/libbpf/src/libbpf.o" "-c" "libbpf/src/libbpf.c"
  cargo:warning=In file included from libbpf/src/libbpf.c:44:
[...]

PR #132, opened 2021-03-31
Multi distro build test

rhdxmr:
I come up with this PR when I ran into the build failure of #131. Because I
tried building RedBPF only under my Pop OS, I could not recognize the build failures
derived from other distros.

And also by `Dockerfile-alpine-20210212` this PR suggests the solution of #130.


rsdy:
I think it would be better to make these part of the CI process. I'm currently trying to break free of CircleCI on the https://github.com/ingraind/ingraind-build/pull/1 PR, and use that to build the base builder images for the 3 main distros.

rhdxmr:
I am not familiar with CircleCI yet, so I have a question about it.
Can I use circleCI readily in localhost to test multi distro build in person?
If some build failure happen, it is required to run build test manually, repeatedly, readily.

rsdy:
In theory this exists, but I could never get it working reliably. It comes with a lot of limitations, and I don't think we can actually use their local build tool for RedBPF.

[CircleCI CLI](https://circleci.com/docs/2.0/local-cli/) has some docs on usage if you would like to try regardless.

My usual workflow involves triggering an actual CI run in the cloud, and working through it on a branch. Not the most fun, but I think you have a lot of scaffolding to play with at this stage.

rhdxmr:
@rsdy 
I added a github action to build redbpf on ubuntu-20.04 and fedora-rawhide containers which are made by `foniod/build-images`.
What about using github action instead of circle CI?
I think it is better because `foniod/build-images` uses it already.

rsdy:
I think it's a good direction! GitHub Actions also looks like a better long-term option than CircleCI, so happy to go down this route!

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,63 @@
...
+        lsb_release -a
+    - name: Initialize git submodules
+      run: |
+        git submodule update --init --recursive
+    - name: Run RedBPF build on ubuntu-20.04 container
+      run: |
+        docker run --privileged \
+        -v /sys/kernel/debug:/sys/kernel/debug:rw \
+        -v /lib/modules:/lib/modules:ro \
+        -v /usr/src:/usr/src:ro \
```
rsdy, line 35: Do we want to override the containers' kernels with whatever we get from GitHub? I feel like this can potentially make it harder to build for a larger range of kernels.

Going forward, I think it might be helpful to think about building for many different kernel versions to see what we can/cannot support.

Just a suggestion, though, I think it's not a straightforward thing so will be a process to get there.

rhdxmr:
I fixed it not to mount volumes of kernel headers and changed to specify `KERNEL_VERSION` environment variable for containers.

rhdxmr:
I could not add build test for ubuntu-18.04.
Because I ran into compile error.

```
   Compiling bpf-sys v1.3.0 (/build/bpf-sys)
error: No suitable version of LLVM was found system-wide or pointed
       to by LLVM_SYS_110_PREFIX.

       Consider using `llvmenv` to compile an appropriate copy of LLVM, and
       refer to the llvm-sys documentation for more information.

       llvm-sys: https://crates.io/crates/llvm-sys
       llvmenv: https://crates.io/crates/llvmenv
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/llvm-sys-110.0.1/src/lib.rs:486:1
    |
486 | / std::compile_error!(concat!(
487 | |     "No suitable version of LLVM was found system-wide or pointed
488 | |        to by LLVM_SYS_",
489 | |     env!("CARGO_PKG_VERSION_MAJOR"),
...   |
496 | |        llvmenv: https://crates.io/crates/llvmenv"
497 | | ));
    | |___^

error: aborting due to previous error

error: could not compile `llvm-sys`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

And the version of included kernel header is 4.15.0-142-generic. It is too old version.

rsdy:
If you remove the CIrcleCI builds, I can merge this and move on from here. I know how to fix the LLVM_SYS issues, but we also need to upgrade to LLVM 12 as the latest Rust compiler already uses that.

rhdxmr:
CircleCI config is deleted.
Thanks,

PR #133, opened 2021-04-02
Fix array index type as u32

belltoy:
As title, type index type of Array should be u32.

### DIFF
--- a/redbpf-probes/src/maps.rs
+++ b/redbpf-probes/src/maps.rs
@@ -119,7 +119,7 @@ macro_rules! define_array {
                 Self {
                     def: bpf_map_def {
                         type_: $map_type,
-                        key_size: mem::size_of::<i32>() as u32,
+                        key_size: mem::size_of::<u32>() as u32,
                         value_size: mem::size_of::<T>() as u32,
                         max_entries,
                         map_flags: 0,
@@ -130,7 +130,7 @@ macro_rules! define_array {
 
             /// Returns a reference to the value at `index`.
             #[inline]
-            pub fn get(&mut self, index: i32) -> Option<&T> {
+            pub fn get(&mut self, index: u32) -> Option<&T> {
                 unsafe {
                     let value = bpf_map_lookup_elem(
                         &mut self.def as *mut _ as *mut c_void,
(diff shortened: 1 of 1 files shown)


rhdxmr:
Here `i32` is used because underlying `bpf_map_*_elem` functions need `i32`.
I agree with that `u32` looks more appropriate for index, but what if `bpf_map_*_elem` start supporting negative index while we use `u32`?

belltoy:
> Here i32 is used because underlying bpf_map_*_elem functions need i32.

As I know, the underlying `bpf_map_*_elem` functions use `*key` as a `const void *` pointer, in the `BPF_MAP_TYPE_ARRAY` type, the key will be use as `u32`([code](https://github.com/torvalds/linux/blob/master/kernel/bpf/arraymap.c#L162)).

Additionally, there is some description for this ARRAY type of map from [bpf(2) manpage](https://man7.org/linux/man-pages/man2/bpf.2.html):

_The key is an array index, and must be exactly bytes._ 


> I agree with that u32 looks more appropriate for index, but what if bpf_map_*_elem start supporting negative index while we use u32?

I can't agree that :)  Who will use a negative integer as index, in C code. Event then, Rust has its high-level abstraction to solve that, like `Index` trait.

belltoy:
It this PR is acceptable, I have to add user space changes as well.

rsdy (approved):
Looks sensible to me. We should really revive the ARM64 test harness to check if these changes pass there, too.

rhdxmr:
> As I know, the underlying bpf_map_*_elem functions use *key as a const void * pointer, in the BPF_MAP_TYPE_ARRAY type, the key will be use as u32(code).

My bad. You are right, u32 is used by linux kernel internally.

I wrote the Array and PerCpuArray a month ago.
When I designed the Array structure I referred to BPF_ARRAY of BCC.
https://github.com/iovisor/bcc/blob/7fd8cad30f910af0470e00130817442091376ab3/src/cc/export/helpers.h#L277
Maybe I was confused by this and I deluded myself that key-type of array map is
32-bit signed integer in linux kernel.

That's why I told you so in the previous comment.

> Rust has its high-level abstraction to solve that, like Index trait.

I also considered `Index` trait, but I could not make use of it because the
value should implement `Copy` trait.

> Who will use a negative integer as index, in C code

Oops, it hurts. In C, I've used to utilize negative index when I dealt with arrays by util functions..



Anyways, This PR causes public api change, so this incurs semver incompatibility.

I am curious that redbpf library has strictly conformed to semver compatibility
rule so that the small api change is allowable without upgrade major version.
Fortunately, the function is not old so the change does not affect much.

rsdy:
@rhdxmr since the change hasn't gone out in a release, technically this is not breaking semver backwards compatibility. I think we can create a new release of all this incoming stuff after the CI is in order.

PR #134, opened 2021-04-04
Fix array index type as u32 for user space APIs

belltoy:
metioned in https://github.com/ingraind/redbpf/pull/133#issuecomment-812864543


rsdy:
@belltoy Please do a force push with a Sign-off-by commit before I can merge this.

PR #137, opened 2021-04-13
allow empty name for maps section

Escapingbug:
ALERT: I don't fully understand the inner problem about this. This problem is found by parsing [tracee compiled ebpf program](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.bpf.c) by redbpf. This is where it causes the error.

Once again, I'M NOT THE EXPERT TOTALLY UNDERSTAND WHY THIS HAPPENS. Please make sure this is the right patch.

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -624,7 +624,7 @@ impl Module {
 
             let section_type = shdr.sh_type;
             let content = data(&bytes, &shdr);
-
+            
             match (section_type, kind, name) {
                 (hdr::SHT_REL, _, _) => add_relocation(&mut rels, shndx, &shdr, shdr_relocs),
                 (hdr::SHT_PROGBITS, Some("version"), _) => version = get_version(&content),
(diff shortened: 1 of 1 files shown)


Escapingbug:
I got some new idea. The problem is not only caused by empty maps name. I'll check what is the perfect approach of fixing this.

rsdy:
From what I understand, we're not completely libbpf compatible, as they put all maps in a single section, while redbpf prefers separate sections for different maps. However, I have not looked into a solution in depth so far, so if you have a good idea how to make the redbpf loader more compatible, that would be helpful!

Escapingbug:
> From what I understand, we're not completely libbpf compatible, as they put all maps in a single section, while redbpf prefers separate sections for different maps. However, I have not looked into a solution in depth so far, so if you have a good idea how to make the redbpf loader more compatible, that would be helpful!

Check #138 , it is a working example for me.

PR #138, opened 2021-04-14
RFC: support load EBPF program written in C

Escapingbug:
Durng my usage of redbpf loading C program, I found a issue that prevents me from successfully load the ebpf program into kernel. After some digging, I found that it is casued by different "maps" section usage protocol.

In C program, the section "maps" contains all maps definitions without actual names of the map. Also, the symbols are used to mark out those maps, with the `st_value` determines the offset into the `maps` section. However, current implementation assumes the `shndx` (section header index) can be used to find the individual map.

This PR is a temp fix of this problem. With this PR, the C program is loadable now (tested with [this program](https://github.com/aquasecurity/tracee/blob/main/tracee-ebpf/tracee/tracee.bpf.c), although I have met several issues I believe not caused by this one. For example, the `kprobe/send_bin` is not attachable, because not found).

I know this is not a clean PR (the error is modified to assist my debugging purpose), I leave this PR here as a RFC. If you guys agree with my primary fixing, I will submit a clean fix of this problem without any other modifications. (The unclear error is another problem, but out of the topic in this PR)

### DIFF
--- a/redbpf/src/error.rs
+++ b/redbpf/src/error.rs
@@ -20,6 +20,7 @@ pub enum Error {
     SymbolNotFound(String),
     ProgramAlreadyLoaded,
     ProgramNotLoaded,
+    ElfError,
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -150,6 +150,7 @@ pub struct StreamVerdict {
     common: ProgramData,
 }
 
+#[derive(Debug)]
 pub struct Map {
     pub name: String,
     pub kind: u32,
@@ -609,12 +610,16 @@ impl SocketFilter {
 impl Module {
     pub fn parse(bytes: &[u8]) -> Result<Module> {
         let object = Elf::parse(&bytes[..])?;
+        let strtab = &object.strtab;
         let symtab = object.syms.to_vec();
         let shdr_relocs = &object.shdr_relocs;
 
         let mut rels = vec![];
         let mut programs = RSHashMap::new();
+        // maps: section header index => map
+        // symval_to_maps: symbol value => map
         let mut maps = RSHashMap::new();
+        let mut symval_to_maps = RSHashMap::new();
 
         let mut license = String::new();
         let mut version = 0u32;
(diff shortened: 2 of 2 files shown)


rhdxmr:
@Escapingbug 
Thank you for your contribution :)

@rsdy
There exists a modification at `pub fn apply` signature.

I have a question about the change.
Breakage of public API is allowable or should we avoid
changing public API by making another function?
What is the principle of redbpf in the case of these modifications?

And I am just curious that `#[allow(dead_code)]` is still required at
`RelocationInfo`.

Escapingbug:
And since you mentioned about `pub fn apply`, why do we make it public? I think the relocation part should be fully internal to redbpf anyway?

rsdy:
@Escapingbug I think this will work well fundamentally, but I'd prefer not breaking API at this point. You're right: in hindsight there's little benefit in exposing the relocation logic directly, however, we chose to do this originally for exposing a flexible interface. 

I think a reasonable tradeoff would be to add a new function that seamlessly chains with the current call:

``` rust
                 rel.apply(&mut programs, &maps, &symtab)
                    .or_else(|_| rel.apply_with_symmap(&mut programs, &symval_to_maps, &symtab))?;
```

This will also make that if/else logic inside `apply()` somewhat cleaner, since they're essentially separate functionalities anyway.

rsdy (changes requested):
@Escapingbug I think this will work well fundamentally, but I'd prefer not breaking API at this point. You're right: in hindsight there's little benefit in exposing the relocation logic directly, however, we chose to do this originally for exposing a flexible interface. 

I think a reasonable tradeoff would be to add a new function that seamlessly chains with the current call:

``` rust
                 rel.apply(&mut programs, &maps, &symtab)
                    .or_else(|_| rel.apply_with_symmap(&mut programs, &symval_to_maps, &symtab))?;
```

This will also make that if/else logic inside `apply()` somewhat cleaner, since they're essentially separate functionalities anyway.

Apart from that, I think propagating the `fd` through the `Error::BPF` enum variant is probably not something we want to have in this form, as it doesn't really say what `fd` we're talking about. I think this will also break arm64 because most opaque types there are unsigned.

Escapingbug:
I hope this fixed the issues. The fd part is my debugging purpose code, should've restore it before commit. Now it should be set.

rsdy:
@Escapingbug That looks great!

One more thing I'd like to ask you to get this merged is re-pushing the commits with a Sign-off line. You can [read more about the process](https://github.com/foniod/project/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) in the docs.

After the DCO check passes in the CI, I'm happy to merge these changes.

Escapingbug:
Added DCO, squashed the commits.

rsdy (approved):

Issue #102, opened 2020-11-03
Difficulties replicating Ingraind DNS probe


s-newman:
I'm writing an XDP program that will check a packet's data for specific content, and if the content is present, drop the packet and send information about it to userspace for futher processing.

The [Ingraind DNS probe](https://github.com/redsift/ingraind/blob/master/ingraind-probes/src/dns/main.rs) is similar to what I'm trying to accomplish, so I tried building and loading the probe to get started with my development. Unfortunately, the eBPF validator won't let me load it due to an `invalid access to packet`.

I found [someone else having a similar issue with XDP](https://www.spinics.net/lists/xdp-newbies/msg01476.html), but their code (and therefore the suggested fix) is in C.

I've been able to get it working if I don't check the packet's data, but then I have to process it in userland, and I can't drop the packet if it contains the content I'm looking for :(

## Environment

- Rust: 1.47.0
- redbpf: 1.2.0
- Linux: 5.4.0-1028-gcp

## Expected behavior

The XDP program should load and attach normally.

## Actual behavior

`Loader::load()` returns a `LoadError` when loading the program, with the following message:

```
invalid access to packet, off=36 size=1, R6(id=2,off=34,r=0)
R6 offset is outside of the packet
```

The complete output is a bit too long for an issue description, so [here's a gist with the full message](https://gist.github.com/s-newman/5147501cf186af20fa74f34b8b5256bc).

## Code Example

`probes/src/block_content/main.rs`:
```rust
#![no_std]
#![no_main]
use redbpf_probes::xdp::prelude::*;

use arf_probes::block_content::Event;

program!(0xFFFFFFFE, "GPL");

#[map("events")]
static mut events: PerfMap<Event> = PerfMap::with_max_entries(10240);

#[xdp("dns_queries")]
pub fn probe(ctx: XdpContext) -> XdpResult {
    let ip = unsafe { *ctx.ip()? };
    let transport = ctx.transport()?;
    let data = ctx.data()?;

    // DNS is at least 12 bytes
    let header = data.slice(12)?;
    if header[2] >> 3 & 0xF != 0u8 {
        return Ok(XdpAction::Pass);
    }

    let event = Event {
        saddr: ip.saddr,
        daddr: ip.daddr,
        sport: transport.source(),
        dport: transport.dest(),
    };

    unsafe {
        events.insert(
            &ctx,
            &MapData::with_payload(event, data.offset() as u32, ctx.len() as u32),
        )
    }

[...]

PR #139, opened 2021-04-18
changes "blk_account_io_completion" to "blk_account_io_done" for kernel version >= 5.8.0

lwintermelon:
The kernel function "blk_account_io_completion" is not available anymore as attach point of Kprobe as of kernel version 5.8.0, use "blk_account_io_done" instead, see more details in https://github.com/iovisor/bcc/pull/3008
And fix some typos.


lwintermelon:
By changing to use doff bitfield to calculate tcp header length, should fix #102 and https://github.com/foniod/ingraind/issues/302

rsdy:
This is great! Thanks!

rsdy (approved):
Looks great! Thanks!

PR #140, opened 2021-04-25
Bcc removal

### LINKED ISSUE #109: Purpose of libbpf and bcc/libbpf?
First, thanks for these libraries and the work you've put into them! I know I've been posting a bunch of issues :smile:

Would it be possible to add some documentation on the structure of bpf-sys? I've been diving into bpf-sys as well as [libbpf/libbpf-rs](https://github.com/libbpf/libbpf-rs) to get a full feel of BPF code in Rust. 

NB. the biggest headache that exists right now (IMO) is the dichotomy between `iproute2` (`ip`/`tc`) loaders/maps as well as `libbpf`, and how that dichotomy plays out in the various Rust libraries.

While going through the `bpf-sys` I saw that `libbpf` is a submodule, but BCC's `libbpf` is also included as just a copy/paste import however upstream BCC uses a submodule for its `libbpf` as well. Why not just use the `libbpf` submodule from upstream?

### LINKED ISSUE #130: Error building against musl

rhdxmr:
- This resolves #109
- Remove bpf-sys/bcc
- Upgrade libbpf version that gets rid of reallocarray so that redbpf can be built with musl (It helps resolve #130)
- Modify redbpf/src/{lib.rs,perf.rs} to eliminate calls of bcc API
- Implement some lower level functions that is not provided by libbpf
- Checked that kprobe, uprobe, socket filter, sockmap examples(all example programs I wrote) work correctly.
- Checked that xdp tool (redbpf-tcp-knock) works correctly
- attach_tracepoint is not tested yet. Because there is no way to run tracepoint program yet.


redbpf/src/lib.rs
```diff
@@ -351,31 +356,37 @@ impl Program {
...
+        attr.name = cname.as_ptr();
+        attr.insns = self.data().code.as_ptr();
+        attr.insns_cnt = self.data().code.len() as u64;
+        attr.license = clicense.as_ptr();
+        attr.__bindgen_anon_1.kern_version = kernel_version;
+        attr.log_level = 1;
+
+        unsafe {
+            let buf_size = 64 * 1024;
+            let log_buffer: MutDataPtr = libc::malloc(buf_size) as MutDataPtr;
```
rsdy, line 377: I'm not a huge fan of `malloc` call here, especially since we free the memory area after the call anyway.

I think a neater option would be using a `[u8; buf_size]` array, and the raw pointer for that. Is there any non-trivial issue why that wouldn't work?

redbpf/src/lib.rs
```diff
@@ -351,31 +356,37 @@ impl Program {
...
+            let fd = bpf_sys::bpf_load_program_xattr(&attr, log_buffer, buf_size as u64);
+            if fd < 0 {
+                let cstr = CStr::from_ptr(log_buffer);
+                eprintln!("error loading BPF program {}", cstr.to_str().unwrap());
+                libc::free(log_buffer as *mut _);
+                Err(Error::BPF)
+            } else {
+                // free should be called to prevent memory leakage
+                libc::free(log_buffer as *mut _);
```
rsdy, line 386: Actually this is free-ing the log buffer of the program, which _may_ cause issues. I'd need to dig deeper to find the exact implications of this, but it might be worth keeping this buffer around in an internal struct somewhere. Have you checked if this is safe to do?

redbpf/src/lib.rs
```diff
@@ -550,10 +524,10 @@ impl XDP {
     pub fn attach_xdp(&mut self, interface: &str, flags: xdp::Flags) -> Result<()> {
         let fd = self.common.fd.ok_or(Error::ProgramNotLoaded)?;
         self.interfaces.push(interface.to_string());
-        let ciface = CString::new(interface).unwrap();
-        let res = unsafe { bpf_sys::bpf_attach_xdp(ciface.as_ptr(), fd, flags as u32) };
-
-        if res < 0 {
+        if let Err(e) = unsafe { attach_xdp(interface, fd, flags as u32) } {
+            if let Error::IO(oserr) = e {
+                eprintln!("error attaching xdp to interface {}: {}", interface, oserr);
```
rsdy, line 529: Is there another instance where we log to the console directly from the library? I think this can be a pretty controversial decision, and I'd generally prefer not introducing this pattern unless we already do it somewhere. A better option might be using `tracing`, which requires manual setup and can be filtered by the host app.

rsdy (changes requested):
This is pretty amazing! Thank you!

I have pretty minor questions and suggestions, overall I really like this patchset, and it's a _much_ needed improvement!

redbpf/src/lib.rs
rhdxmr: I never considered that the output buffer kept being used by something inside libbpf after the `bpf_load_program_xattr` returns.
If it really is, the API is bug prone and there should be a lot of noticeable CAUTIONS around the function.

I'll check this is safe too. Thanks

redbpf/src/lib.rs
rhdxmr: I was focusing on eliminating bcc and I tried to avoid touching anything else.
malloc has been used there and I just did not change it. ;)
OK, I will replace malloc with an array.

redbpf/src/lib.rs
rhdxmr: Good point. I will fix this. Thank you :+1:

rhdxmr: Now `error!` of `tracing` crate is used instead of `eprintln!`

redbpf/src/lib.rs
rhdxmr: I checked that `log_buffer` is not used by libbpf. It is merely passed to `bpf` system call.

Inside linux kernel, `log_buf` of `bpf_attr` is used by BPF verifier and the buffer is not maintained by kernel after `bpf_check` is finished.

The call stack I checked is this
```
bpf_verifier_vlog (kernel/bpf/verifier.c)
bpf_check (kernel/bpf/verifier.c)
bpf_prog_load (kernel/bpf/syscall.c)
```

So I think it is fine to free `log_buffer` after `bpf_load_program_xattr` ends.

redbpf/src/lib.rs
rhdxmr: `malloc` is removed and `Vec.as_mut_ptr` is called to provide log buffer.

rhdxmr:
I rebased and re-pushed this PR

rsdy (approved):
Looks good to me! Thanks!

PR #142, opened 2021-05-05
Improve kernel tree handling to support NixOS

rsdy:
This must remain backwards compatible while also making the trees we accept from the `KERNEL_SOURCE` environment variable more versatile. This patchset allows building on NixOS.


rhdxmr (approved):
LGTM :+1:

PR #148, opened 2021-05-15
Add map pin

rhdxmr:
- Add map pinning feature (`Map::pin`, `Map::unpin` and `Map::from_pin_path`)
- Detach perf event and close its RawFd when `KProbe`/`UProbe` is dropped
- Close RawFd of `ProgramData` when it is dropped
- Close RawFd of `Map` when it is dropped
- Add accessors to `Loaded` and `Module` to get a reference of structure by name
- Rename `Loaded::stream_parser` to `Loaded::stream_parsers` (plural)
- Fix trivial bug of build script

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -741,6 +741,10 @@ impl Module {
         self.programs.iter().find(|p| p.name() == name)
     }
 
+    pub fn program_mut(&mut self, name: &str) -> Option<&mut Program> {
+        self.programs.iter_mut().find(|p| p.name() == name)
+    }
+
     pub fn kprobes(&self) -> impl Iterator<Item = &KProbe> {
         use Program::*;
         self.programs.iter().filter_map(|prog| match prog {
@@ -755,6 +759,10 @@ impl Module {
             KProbe(p) | KRetProbe(p) => Some(p),
             _ => None,
         })
+    }
+
+    pub fn kprobe_mut(&mut self, name: &str) -> Option<&mut KProbe> {
+        self.kprobes_mut().find(|p| p.common.name == name)
     }
 
     pub fn uprobes(&self) -> impl Iterator<Item = &UProbe> {
@@ -773,6 +781,10 @@ impl Module {
         })
     }
 
+    pub fn uprobe_mut(&mut self, name: &str) -> Option<&mut UProbe> {
+        self.uprobes_mut().find(|p| p.common.name == name)
+    }
+
     pub fn xdps(&self) -> impl Iterator<Item = &XDP> {
         use Program::*;
         self.programs.iter().filter_map(|prog| match prog {
@@ -787,6 +799,10 @@ impl Module {
             XDP(p) => Some(p),
             _ => None,
         })
+    }
+
+    pub fn xdp_mut(&mut self, name: &str) -> Option<&mut XDP> {
+        self.xdps_mut().find(|p| p.common.name == name)
     }
 
     pub fn socket_filters(&self) -> impl Iterator<Item = &SocketFilter> {
@@ -805,6 +821,10 @@ impl Module {
         })
     }
 
+    pub fn socket_filter_mut(&mut self, name: &str) -> Option<&mut SocketFilter> {
+        self.socket_filters_mut().find(|p| p.common.name == name)
+    }
+
     pub fn trace_points(&self) -> impl Iterator<Item = &TracePoint> {
         use Program::*;
         self.programs.iter().filter_map(|prog| match prog {
@@ -821,7 +841,11 @@ impl Module {
         })
     }
 
-    pub fn stream_parser(&self) -> impl Iterator<Item = &StreamParser> {
+    pub fn trace_point_mut(&mut self, name: &str) -> Option<&mut TracePoint> {
+        self.trace_points_mut().find(|p| p.common.name == name)
+    }
+
+    pub fn stream_parsers(&self) -> impl Iterator<Item = &StreamParser> {
         use Program::*;
         self.programs.iter().filter_map(|prog| match prog {
             StreamParser(p) => Some(p),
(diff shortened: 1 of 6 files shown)


redbpf/src/lib.rs
```diff
@@ -392,6 +414,61 @@ impl Program {
...
+    fn drop(&mut self) {
+        if self.fd.is_some() {
+            unsafe {
+                let _ = libc::close(self.fd.unwrap());
+            }
+        }
+    }
+}
+
+fn pin_bpf_obj<P: AsRef<Path>>(fd: RawFd, pathname: P) -> Result<()> {
```
rsdy, line 427: You could also use the `pathname: impl AsRef<Path>` syntax here, that's a bit more modern.

redbpf/src/lib.rs
```diff
@@ -392,6 +414,61 @@ impl Program {
...
+    let parent = pathname.as_ref().parent();
+    if parent.is_none() {
```
rsdy, line 429: This happens when `pathname` is `/`? Might be clearer to use `file` as argument name, and potentially a better error than `ErrorKind::InvalidInput`, which is not super descriptive. It might also be worth validating if this is on a bpf file system path based on /proc/mounts for instance.

redbpf/src/lib.rs
```diff
@@ -415,10 +492,45 @@ impl KProbe {
...
+    ///     kprobe.attach_kprobe(&kprobe.name(), 0).unwrap();
+    ///     // do some stuff...
+    ///     kprobe.detach_kprobe(&kprobe.name(), 0).unwrap();
+    /// }
+    /// ```
+    pub fn detach_kprobe(&mut self, fn_name: &str, offset: u64) -> Result<()> {
+        // bpf program is detached from perf event and the perf event is closed
+        // by dropping KProbeAttachmentPoint
+        self.attachment_points
+            .retain(|ap| ap.fn_name != fn_name || ap.offset != offset);
```
rsdy, line 530: Might be more readable to use the De Morgan's laws here as `!(ap.fn_name == fn_name && ap.offset == offset)`

redbpf/src/lib.rs
```diff
@@ -485,10 +597,56 @@ impl UProbe {
...
+        &mut self,
+        fn_name: Option<&str>,
+        offset: u64,
+        target: &str,
+        pid: Option<pid_t>,
+    ) -> Result<()> {
+        // bpf program is detached from perf event and the perf event is closed
+        // by dropping UProbeAttachmentPoint
+        self.attachment_points.retain(|ap| {
+            ap.fn_name.as_ref().and_then(|x| Some(x.as_str())) != fn_name
```
rsdy, line 642: `and_then(|x| Some(...))` is probably nicer as a `map(...)`.

rsdy (changes requested):
I don't entirely understand the semantics of a few functions, and how the `Option<>` types would work with attaching and detaching to a function. The rest of the comments are more pedantic, and I'm not too fussy about.

redbpf/src/lib.rs
rhdxmr: Ok. `map` looks more concise.

redbpf/src/lib.rs
rhdxmr: Oh I see. I didn't know which is modern or old style because I started learning rust with rust 2018. I just referred the code written in `std::path::Path` structure.
I'll reflect your suggestion.

redbpf/src/lib.rs
```diff
@@ -485,10 +597,56 @@ impl UProbe {
...
+    /// use redbpf::Module;
+    /// let mut module = Module::parse(&std::fs::read("file.elf").unwrap()).unwrap();
+    /// let uprobe = module.uprobe_mut("count_strlen").expect("bpf program not found");
+    /// uprobe.attach_uprobe(Some(&uprobe.name()), 0, "/lib/x86_64-linux-gnu/libc-2.30.so", None).unwrap();
+    /// // do some stuff...
+    /// uprobe.detach_uprobe(Some(&uprobe.name()), 0, "/lib/x86_64-linux-gnu/libc-2.30.so", None);
+    /// ```
+    pub fn detach_uprobe(
+        &mut self,
+        fn_name: Option<&str>,
```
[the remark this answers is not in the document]
rhdxmr: `detach_uprobe` has the same signature of `attach_uprobe`.
The intention is that `detach_uprobe` can be called by passing the exactly same arguments which were passed to `attach_uprobe`.

For example, if `attach_uprobe(None, 1000, "libc.so", None)` is called, the uprobe event will be detached by calling `detach_uprobe(None, 1000, "libc.so", None)`.

redbpf/src/lib.rs
```diff
@@ -485,10 +597,56 @@ impl UProbe {
                     perf::open_uretprobe_perf_event(&path, offset + sym_offset, pid)?
                 }
             };
-            perf::attach_perf_event(fd, pfd)
+            let ret = perf::attach_perf_event(fd, pfd);
+            if ret.is_ok() {
+                self.attachment_points.push(UProbeAttachmentPoint {
+                    fn_name: fn_name.and_then(|fn_name| Some(fn_name.to_owned())),
```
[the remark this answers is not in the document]
rhdxmr: Thank you for this advice. I couldn't come up with this.

redbpf/src/lib.rs
rhdxmr: Yes. If `pathname` is `/`, `ErrorKind::InvalidInput` is returned. I will think of other error value. Any suggestions?

In the first patch, if `pathname` is not valid bpf fs, error is returned from `bpf_sys::bpf_obj_pin`
`IO(Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" })`
But it just gives small information to users. And it is difficult to understand root cause.

Validation will help users correct error.

rhdxmr:
I modified the code and pushed it again

rsdy (approved):

PR #149, opened 2021-05-23
Use `geteuid` instead of `getuid` for privilege checks

p-e-w:
Privilege boundary enforcement in the kernel relies on the *effective* user ID, so userspace code preparing to invoke privileged kernel functionality should always check for the eUID.

This matters especially for `setuid`, which previously was not working with the RedBPF tools:

```
$ sudo chown root redbpf-iotop
$ sudo chmod ug+s redbpf-iotop
$ ./redbpf-iotop
redbpf-iotop: You must be root to use eBPF!
```

With this change, it works as expected, and correctly recognizes that the effective privileges are that of the root user.


rhdxmr (approved):
It looks better to use euid instead of uid.
Thank you for your contribution :+1:

PR #151, opened 2021-06-05
Share maps between independent BPF programs

rhdxmr:
- Add #[map(from_pin_file = "...")] meta into map attribute macro
- Add mapxattr/maps/<item_name> sections only if the new meta is used
- Parse mapxattr/maps/<item_name> sections and let programs share already existing maps
- Add sharemap1/sharemap2 examples
- Fix example probe compile warnings
- Substitute ingraind with foniod

### DIFF
--- a/redbpf-macros/src/lib.rs
+++ b/redbpf-macros/src/lib.rs
@@ -55,8 +55,8 @@ use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
 use syn::token::Comma;
 use syn::{
-    parse_macro_input, parse_quote, parse_str, Expr, ExprLit, File, ItemFn, ItemStatic, Lit, Meta,
-    Result,
+    parse_macro_input, parse_quote, parse_str, AttributeArgs, Expr, ExprLit, File, ItemFn,
+    ItemStatic, Lit, Meta, NestedMeta, Result,
 };
 
 fn inline_string_literal(e: &Expr) -> (TokenStream2, TokenStream2) {
@@ -156,7 +156,7 @@ pub fn impl_network_buffer_array(_: TokenStream) -> TokenStream {
 }
 
 /// Attribute macro that must be used when creating [eBPF
-/// maps](https://ingraind.org/api/redbpf_probes/maps/index.html).
+/// maps](../../redbpf_probes/maps/index.html).
 ///
 /// The default `#[map]` places the map into a section of the resulting
 /// ELF binary called `maps/<item_name>`.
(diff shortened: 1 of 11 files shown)


rhdxmr:
@rsdy Can you review this PR?

rsdy:
Sorry for the delay on this. I'm on holidays, but will be able to do a proper review this weekend.

redbpf-macros/src/lib.rs
```diff
@@ -168,6 +168,16 @@ pub fn impl_network_buffer_array(_: TokenStream) -> TokenStream {
...
 /// been deprecated in favor of `#[map]` or `#[map(link_section = "maps/foo")]`
 ///
+/// If you wish to share maps which already exist and are pinned to files, use
+/// `#[map(from_pin_file = "/path/to/pin/file")]`. This adds
+/// `mapxattr/maps/<item_name>` section to the resulting ELF binary. When the
+/// section is parsed by
+/// [redbpf::Module::parse](../../redbpf/struct.Module.html), redbpf makes BPF
+/// programs share the pinned map.
+///
+/// **CAUTION:** There should exist the designated pin files before the
```
rsdy, line 178: I'm not sure this is a great idea, as it's a bit of a catch 22. It might be best putting some fool-proof logic in there to make sure the error doesn't bubble up and cause hard to explain behaviour when trying to run a program.

rsdy (approved):
This is really cool! Definitely a large piece of work. A few notes to keep in mind for the future:
 * most of the changeset is reformatting and rewriting comments that would be nicer in a separate PR, as they're unrelated
 *  please let me know if you have an idea for a a fail-proof API around this.

In addition, please create a separate PR that adds your name and email to the maintainers list of the crates, as I've seen you didn't do this while changing my email.

redbpf-macros/src/lib.rs
rhdxmr: I agree with that this API is easy to fail. I will try figuring out how to enhance the API.

At first I considered to provide userspace API only, because it is intuitive and easy to find where the root cause is..

But it seemed hard to use that API and it seemed complicated to understand that unless new design of Module is introduced.. So I chose to offer attribute meta instead. It is easier to understand and simple to use.

I think that it would be better to split the Module::parse() into parse() and load(). And it allows users to do something between parse() and load(). But changing API like this would bring big difference to users and it really would bother them so I could not choose that. Maybe I can add other API without touching the original, but wouldn't it be better to do that when major version upgrade?

rhdxmr:
Hello @rsdy

`from_pin_file` meta of `map` attribute macro has been retracted because I
could not come up with enhancing fail proof API of it.
And its relevant code for handling `mapxattr/` ELF sections is also discarded.

Instead `ModuleBuilder` is introduced to support separation of parsing and
loading BPF maps procedures. `ModuleBuilder` is a new structure of userspace
API and it offers `ModuleBuilder::replace_map` method which replaces parsed BPF
map with the map passed by users. And that method is intended to be used in
combination with `Map::from_pin_file`.

Substitution of links in comment will be sent as a different PR later.

rsdy (approved):

rsdy:
I like this approacha a lot better, az it looks much simpler on both sides to use, and has fewer moving pieces. I'm happy to go with this unless you have any final changes you'd like to make?

rhdxmr:
I reconsidered to figure out that any changes are needed. But I could not come up with any ideas so I am going to merge this. Thanks,

PR #156, opened 2021-07-17
Fix dependency versions at Cargo.toml of examples

rhdxmr:
Since cargo v1.54, invalid syntax for version requirements is rejected
cf)
https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-154-2021-07-29
https://github.com/rust-lang/cargo/pull/9508


rsdy:
I think it would be better off using `"*"` here so we don't need to bump it all the time.

rhdxmr:
> I think it would be better off using `"*"` here so we don't need to bump it all the time.

That's a better suggestion. I will change it. Thanks

rsdy (approved):

PR #159, opened 2021-08-01
Vmlinux

rhdxmr:
- Upgrade libbpf submodule
- Build libbpf static library using `make`
- Upgrade `bindgen` crate
- Add `bpf_sys::type_gen::VmlinuxBtfDump` that generates vmlinux.h from vmlinux image
- Add logic on how to select methods of generating rust bindings

> # Methods of generating rust bindings
> 
> Rust bindings for structs and enums of the Linux kernel are used by BPF programs.
> There are two methods for generating the rust bindings.
> 
> 1. Generate rust bindings from the Linux kernel headers that are pre-installed
> in the system. The Linux kernel headers are located by
> [`bpf_sys::headers`](../../bpf_sys/headers/index.html).
> 
> 2. Generate rust bindings from `vmlinux.h`. It is generated on the spot by
> [`bpf_sys::type_gen`](../../bpf_sys/type_gen/index.html). The vmlinux is an
> image of the Linux kernel so all data types including internal structs and
> enums can be dumped into C source code from the vmlinux image.
> 
> Users of redBPF can select a preferred method for generating rust bindings by
> setting environment variables explained as below.
> 
> # Rules about method selection
> 
> In order to select a method between two methods, three environment variables are
> involved: `KERNEL_SOURCE`, `KERNEL_VERSION` and `REDBPF_VMLINUX`.
> 
> Case 1. No setting
> 
> If none of the three environment variables are set, both two methods of
> generating rust bindings will be tried. First, the method with the Linux kernel
> headers is tried. If it fails, then the method with vmlinux is tried as
> fallback.
> 
> Case 2. When `REDBPF_VMLINUX` is set
> 
> `REDBPF_VMLINUX` takes precedence over `KERNEL_SOURCE` and `KERNEL_VERSION`
> environment variables. So the method of generating rust bindings of the Linux
> kernel data structures for BPF programs is to vmlinux.h that is generated from
> vmlinux image. No pre-installed kernel headers are required in this case.
> 
> Case 3. `REDBPF_VMLINUX` is not set, but any of `KERNEL_SOURCE` or
> `KERNEL_VERSION` environment variables are set
> 
> The method of generating rust bindings of the Linux kernel data structures for
> BPF programs is to use the kernel headers in the specified path. In this case
> vmlinux is not required at all.
> 
> # Possible `REDBPF_VMLINUX` values
> 
> 1. A path to the custom vmlinux. e.g) `REDBPF_VMLINUX=/boot/my-vmlinux-5.11.0`
> 
> 2. Special treatment for `REDBPF_VMLINUX=system`. If `system` is given, redBPF
> tries to probe vmlinux from the well-known system paths and uses it
> 
> 3. `REDBPF_VMLINUX` not set. The behavior of redBPF depends on whether
> `KERNEL_SOURCE` and `KERNEL_VERSION` environment variables are given or not.


bpf-sys/src/type_gen.rs
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        Ok(VmlinuxBtfDump {
+            whitelist: None,
+            btfptr,
+        })
+    }
+
+    /// Add regex `pattern` into whitelist of BTF types
+    pub fn whitelist(mut self, pattern: &str) -> Self {
```
rsdy, line 121: Same here, would be nicer with `allowlist` as terminology.

bpf-sys/src/type_gen.rs
```diff
@@ -0,0 +1,253 @@
...
+        self
+    }
+
+    /// Dump BTF types as C source code into `outfile` including all the
+    /// necessary dependent types.
+    pub fn generate(self, outfile: impl AsRef<Path>) -> Result<()> {
+        let path_str = outfile.as_ref().to_str().ok_or(TypeGenError::InvalidPath)?;
+        let cpath = CString::new(path_str).unwrap();
+        let cmode = CString::new("w").unwrap();
+        let stream = unsafe { libc::fopen(cpath.as_ptr(), cmode.as_ptr()) };
```
rsdy, line 138: I understand this is a ported code from C, but is there a reason you're using C APIs instead of the `std::fs` equivalent from Rust? I feel these would be low hanging fruits, and not particularly hard to adapt.

bpf-sys/src/type_gen.rs
```diff
@@ -0,0 +1,253 @@
...
+impl Drop for VmlinuxBtfDump {
+    fn drop(&mut self) {
+        unsafe {
+            btf__free(self.btfptr);
+        }
+    }
+}
+
+// wrapping vfprintf to get rid of return type
+unsafe extern "C" fn vfprintf_wrapper(
```
rsdy, line 231: I feel like this probably can go as well, if the code above uses `std::fs` instead of the `libc` api?

rsdy (changes requested):

bpf-sys/src/type_gen.rs
rhdxmr: https://github.com/libbpf/libbpf/blob/506a544834573905ada61da8e00f54b04f0caf43/src/btf.h#L152

libbpf requires `btf_dump_printf_fn_t` so I defined `vfprintf_wrapper` that
directly calls `vfprintf`. Without `vfprintf`, `vfprintf_wrapper` should deal
with `va_list`. But I felt it is tricky.

And `vfprintf` requires `FILE*` so I provided it by calling `libc::fopen`.

I wanted to use `std::fs` but I din't know whether it provides a function returning `FILE *`.

bpf-sys/src/type_gen.rs
rhdxmr: Good. I will rename this.

rhdxmr:
@rsdy I renamed `VmlinuxBtfDump::whitelist` to `VmlinuxBtfDump::allowlist` and re-pushed.

But I could not figure out how to use `std::fs` instead of `libc::fopen` because of the reason I mentioned above. Do you have ideas?

bpf-sys/src/type_gen.rs
rsdy: [`File::into_raw_fd()`](https://doc.rust-lang.org/std/fs/struct.File.html#method.into_raw_fd) helps you get the underlying fd, and you can use `vdprintf` with that fd to achieve the same results. There's also a [crate](https://crates.io/crates/va_list) that could help with dealing with variadic arguments, so you can keep most of this purely rust (e.g. get the raw pointer (_not_ fd) to the `File` instance if you know it outlives the callback, and convert it back on the other side with `unsafe`, then do something rusty to write the data to the file).

rhdxmr: It is a good idea. I will try to fix it. Thanks,

rhdxmr:
@rsdy I removed calling `libc::fopen` and used `std::fs::File` instead. It looks better. Thanks,

bpf-sys/src/type_gen.rs
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                }
+                if btf_dump__dump_type(dumpptr.0, type_id) < 0 {
+                    return Err(TypeGenError::DumpError);
+                }
+            }
+        }
+
+        let mut fobj = unsafe { File::from_raw_fd(rawfd.0) };
+        // disable dropping `RawFdWrapper` because `File` takes control of `RawFd`
```
rsdy, line 199: If this is the case, do we still need the `impl Drop for RawFdWrapper`?

bpf-sys/src/type_gen.rs
```diff
@@ -0,0 +1,240 @@
...
+                        continue;
+                    }
+                }
+                if btf_dump__dump_type(dumpptr.0, type_id) < 0 {
+                    return Err(TypeGenError::DumpError);
+                }
+            }
+        }
+
+        let mut fobj = unsafe { File::from_raw_fd(rawfd.0) };
```
rsdy, line 198: Maybe we could do an `impl Into<File> for RawFdWrapper`, so there's RAII continuity between the two types the other way around. `let fobj: File = rawfd.into()` would then consume `rawfd`.

bpf-sys/src/type_gen.rs
rhdxmr: Yes, `Drop` is needed because `generate` method can early escape before this line.

bpf-sys/src/type_gen.rs
rhdxmr: Cool. Using `Into` trait looks elegant. I will try it.

rsdy (approved):

PR #160, opened 2021-08-04
Add llvm11 and llvm12 feature

rhdxmr:
The latest Alpine Linux does not support LLVM 12 yet so use LLVM 11 if target env is `musl`

### DIFF
--- a/cargo-bpf/src/llvm.rs
+++ b/cargo-bpf/src/llvm.rs
@@ -1,3 +1,15 @@
+#[cfg(feature = "llvm-sys-110")]
+use llvm_sys_110 as llvm_sys;
+#[cfg(feature = "llvm-sys-120")]
+use llvm_sys_120 as llvm_sys;
+
+cfg_if::cfg_if! {
+    if #[cfg(feature = "llvm-sys-110")] {
+        #[rustversion::since(1.52)]
+        compile_error!("Can not use LLVM11 with Rust >= 1.52");
+    }
+}
+
 use anyhow::{anyhow, Result};
 use llvm_sys::bit_writer::LLVMWriteBitcodeToFile;
 use llvm_sys::core::*;
@@ -139,9 +151,6 @@ pub unsafe fn process_ir(context: LLVMContextRef, module: LLVMModuleRef) -> Resu
         }
         func = LLVMGetNextFunction(func);
     }
-
-    // This removes .BTF sections. Do not call it.
-    // LLVMStripModuleDebugInfo(module);
 
     Ok(())
 }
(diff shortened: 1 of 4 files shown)


redbpf/src/lib.rs
```diff
@@ -498,7 +499,7 @@ fn pin_bpf_obj(fd: RawFd, file: impl AsRef<Path>) -> Result<()> {
             error!("error on statfs {:?}: {}", path, io::Error::last_os_error());
             return Err(Error::IO(io::Error::last_os_error()));
         }
-        if stat.f_type != libc::BPF_FS_MAGIC {
+        if stat.f_type as i64 != BPF_FS_MAGIC {
```
rhdxmr, line 502: In musl target environment, there is no `libc::BPF_FS_MAGIC`.

rsdy (changes requested):
I'm not sure about this. This essentially forces us to always compile with LLVM11 even if the host does support LLVM12 but we want to use musl for some reason.

I think a better approach would be to introduce some sort of a feature flag that allows us to use LLVM11 during build. I've not done this, but I think a similar approach would work. A note in the README and doc should then help navigating this. In any case, we should default to LLVM12 since that is what Rust targets right now.

rhdxmr:
Ok. I understand that.
How about passing LLVM version string via cfg or feature option?

rsdy:
That could possibly work. Or we could just ship both a `llvm11` and a `llvm12` feature, and default to `llvm12`

rhdxmr:
@rsdy Features `llvm11` and `llvm12` are added. LLVM12 is default and LLVM11 is only used if it is specified.

rhdxmr:
oh no, I must miss something.

cargo-bpf/src/llvm.rs
```diff
@@ -1,3 +1,14 @@
+cfg_if::cfg_if! {
+    if #[cfg(feature = "llvm11")] {
+        use llvm_sys_110 as llvm_sys;
+        #[rustversion::since(1.52)]
+        compile_error!("Can not use LLVM11 with Rust >= 1.52");
```
rsdy, line 5: 👍

rsdy (approved):

cargo-bpf/src/llvm.rs
```diff
@@ -1,3 +1,14 @@
+cfg_if::cfg_if! {
```
rsdy, line 1: I think this is probably a more robust way of doing this. I do really like that it throws an error if there's an incompatible configuration, though!

```rust
#[cfg(feature = "llvm-sys-110")]
use llvm_sys_110 as llvm_sys;
#[cfg(feature = "llvm-sys-120")]
use llvm_sys_120 as llvm_sys;

if #[cfg(feature = "llvm11")] {
        #[rustversion::since(1.52)]
        compile_error!("Can not use LLVM11 with Rust >= 1.52");
}
```

rsdy (changes requested):

cargo-bpf/src/llvm.rs
rhdxmr: Ok I reflected your suggestion to the code

rsdy:
Looking good! Thanks for sorting this.

PR #163, opened 2021-08-11
Remove version fields from Cargo.toml that are not for `creates.io`

rhdxmr:
Version fields for redbpf packages are removed from `Cargo.toml`. But I only modified packages that are not uploaded to `creates.io`


rsdy:
One of the other issues here is that `ˆ` does not work with `-rc` pre-release version numbers in cargo, So that would also make sense to change in one go.

rsdy (approved):

PR #169, opened 2021-08-22
Change helper return values from i32 to i64

### LINKED ISSUE #145: Change return type of functions which expose return of bpf helpers
Since `6f8e021c3c8904d19b659898e4b8b83782ef62bb` of libbpf changed return type of bpf helpers from `int` to `long`, there exist inconsistency of return type where functions directly return the value from bpf helpers.
But changing return type of public API incurs semver breakage, and this fix will be deferred until next major version upgrade.

List of functions which are needed to be fixed:
- bpf_probe_read
- StackTrace.stack_id
- ProgramArray.bpf_tail_call

rhdxmr:
Follow up libbpf's 6f8e021c commit

This resolves #145 

- redbpf_probes::helpers::bpf_probe_read
- redbpf_probes::helpers::bpf_perf_event_output
- redbpf_probes::maps::StackTrace::stack_id
- redbpf_probes::maps::ProgramArray::tail_call

and 
- redbpf::StackTrace::get, delete

### DIFF
--- a/examples/example-probes/src/mallocstacks/main.rs
+++ b/examples/example-probes/src/mallocstacks/main.rs
@@ -15,7 +15,6 @@ static mut malloc_event: PerfMap<MallocEvent> = PerfMap::with_max_entries(1024);
 fn malloc(regs: Registers) {
     let mut mev = MallocEvent {
         stackid: 0,
-        _padding: 0,
         size: regs.parm1(),
     };
 
--- a/examples/example-probes/src/mallocstacks/mod.rs
+++ b/examples/example-probes/src/mallocstacks/mod.rs
@@ -1,7 +1,6 @@
 #[derive(Debug)]
 #[repr(C)]
 pub struct MallocEvent {
-    pub stackid: i32,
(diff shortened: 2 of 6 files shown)


rsdy:
This looks good, albeit i'm slightly concerned changing `c_int` to `i64` indiscriminately because on arm64 a lot of amd64 i64's are u64's in this context.

Since it looks like GitHub Actions can't run ARM64 builders right now without extra magic (https://github.com/actions/virtual-environments/issues/2552), I think we can drop arm64 verification for 2.0 and roll with it.

rsdy (approved):

rhdxmr:
@rsdy 
>  on arm64 a lot of amd64 i64's are u64's in this context.
what does that mean? I am not familiar with arm64 so I don't get it.

rsdy:
A fair amount of places where Linux on x86 uses unsigned types, arm64 uses unsigned ones, and it's not always straightforward, where. I don't think it's practically an issue now, as I don't expect many arm64 users.

Issue #173, opened 2021-08-27
redbpf-probes build script failure


rhdxmr:
The build script of `redbpf-probes` 2.0 exits abnormally with this error.

> error: header '../bpf-sys/libbpf/src/bpf_helpers.h' does not exist.

This happens when the `redbpf-probes` dependency is defined only with a version number.

I found that a copy of `bpf_helpers.h` was removed from `redbpf-probes` by c33138d7de322bb92cf2e690af29ebfa0055038f on May 10.

I could not run into this issue before `redbpf-probes` 2.0 was released to `crates.io`
because everything's ok with the `redbpf-probes` dependency if it is defined by using a `path =` option.

PR #174, opened 2021-08-27
Add redbpf-probes/include/libbpf git sub-module

rhdxmr:
Handles #173


rhdxmr:
@rsdy submodule is discarded and two symlinks are added.
Also I checked that `cargo publish` command follows the symlinks and includes the original files into the crate.

rsdy (approved):

Issue #183, opened 2021-09-14
error: options `-C embed-bitcode=no` and `-C lto` are incompatible


rhdxmr:
With rust v1.55, redbpf-tools compilation fails

```
error: options `-C embed-bitcode=no` and `-C lto` are incompatible
```

But with rust v1.54, compilation succeeds

rhdxmr:
Close this issue. it's been resolved by #184

p-e-w:
I suggest making a patch release with the fix, since this bug completely breaks RedBPF for users of the latest Rust version, with no obvious workaround other than downgrading.

rhdxmr:
Thank you @p-e-w for your suggestion. I agree that it is a sufficient rationale to release a new version of redbpf.

@rsdy As p-e-w suggested, why don't we release a new version for users who use rust 1.55?

rsdy:
Thanks both, I have now pressed release on 2.0.2!

rsdy:
Oops, this is actually not going to go through because of an Alpine-specific issue.

rhdxmr:
@rsdy @p-e-w Thanks to you users can get redbpf 2.0.2 from now on.

PR #184, opened 2021-09-17
Use CARGO_ENCODED_RUSTFLAGS to pass --embed-bitcode=yes to rustc

rhdxmr:
Previously, cargo-bpf used the RUSTFLAGS environment variable to pass
--embed-bitcode=yes to all invocations of rustc. But since rust 1.55,
this behavior has changed. Even if RUSTFLAGS='-C embed-bitcode=yes' is
specified, this flag is not passed to the rustc so that compilation
fails with an error: "error: options `-C embed-bitcode=no` and `-C lto`
are incompatible".

Since cargo v1.55, this error occurs because it passes `-C
embed-bitcode=no` to rustc but RUSTFLAGS='-C embed-bitcode=yes' is not
appended to the rustc flags.

To resolve this problem, since rust v1.55 use CARGO_ENCODED_RUSTFLAGS
instead of RUSTFLAGS to pass --embed-bitcode=yes to rustc.

### DIFF
--- a/cargo-bpf/src/build.rs
+++ b/cargo-bpf/src/build.rs
@@ -71,6 +71,32 @@ impl From<Error> for CommandError {
     }
 }
 
+#[rustversion::since(1.55)]
+fn create_rustflags() -> (String, String) {
+    let mut flags = String::new();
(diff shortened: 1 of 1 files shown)


rhdxmr:
This resolves #183

rsdy (approved):

PR #188, opened 2021-09-24
Disable Alpine until they have LLVM12

rsdy:
Disable Alpine builds on CI until they ship with LLVM12. I don't see how we could simultaneously fix the issue described in #187 and maintain Alpine support, as they still ship with LLVM11 _and_ the newest Rust compiler.

@p-e-w @rhdxmr Please do let me know if you know of a cleaner solution for this.


rhdxmr:
@rsdy @p-e-w 
I'm merging this because I could not come up with a cleaner solution right now.
If I'll get a better solution, then I'll send a PR to re-enable alpine 3.14 build test.

rhdxmr (approved):

rhdxmr:
@rsdy Are ready to roll out a new version? Then let's go

PR #193, opened 2021-09-28
Allow unsafe functions to be used as probes

p-e-w:
Probe code commonly contains lots of unsafe operations (accessing `static mut` maps, dereferencing `*const` values from registers, `mem::transmute` etc). Instead of riddling the code with `unsafe {}` blocks, it can often be cleaner to just mark the entire function as `unsafe`.

Alas, that is currently impossible, because the attribute macros generate code that calls the probe function in a safe context. This PR simply wraps each probe function invocation in an `unsafe` block, thus enabling `unsafe fn`s to be used as probes.

Note that

* having an `unsafe` block that contains only safe statements is perfectly fine in Rust
* unsafe contexts don't propagate down the call chain, so if the function is *not* declared as `unsafe`, the compiler still enforces safety checks
* the `unsafe` blocks added here only contain a function call and a pre-evaluated argument, so no unsafety is actually being introduced.

Thus this change simply adds the capability to use `unsafe fn`s if you want to, without weakening any checks if you don't.

### DIFF
--- a/redbpf-macros/src/lib.rs
+++ b/redbpf-macros/src/lib.rs
@@ -315,7 +315,7 @@ fn wrap_kprobe(item: ItemFn) -> ItemFn {
     parse_quote! {
         fn #outer_ident(ctx: *mut c_void) -> i32 {
             let regs = ::redbpf_probes::registers::Registers::from(ctx);
-            let _ = #ident(regs);
+            let _ = unsafe { #ident(regs) };
             return 0;
 
             #item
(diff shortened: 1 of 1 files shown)


rsdy (approved):
This looks to me like a safe change, and a well reasoned PR. I'm keen to get this in, but just to double check, please have a look @rhdxmr .

rhdxmr:
Hello @p-e-w

I reviewed your PR and I am fully satisfied with your code.
Thanks for this modification and this change is valuable.
And I felt you made this change upon mature consideration. I felt that from
your elaborate description.

I am merging this PR.

Thank you.

rhdxmr (approved):

Issue #150, opened 2021-05-23
Document which Rust types can be placed into eBPF maps


p-e-w:
It appears that the RedBPF toolchain imposes currently undocumented restrictions on map value types that go beyond those expressed by the generic constraints:

* All variants of an enum must have the same size in memory.
* Struct fields are (mostly?) required to be aligned to multiples of 32 bits, though sometimes 64-bit alignment is required, and some structs don't work at all despite their fields being aligned.
* Tweaking alignments with `#[repr(align(n))]` has no effect whatsoever. Whether or not `#[repr(C)]` is used doesn't seem to matter either.
* Double-width types like `u128` cannot be placed in nested structs. They sometimes work in flat structs though.
* Fields with array types are hit-and-miss, and whether they work or not appears to depend on the phase of the moon.
* Getting any of this wrong results in an incomprehensible error message from libbpf, which gives no hint whatsoever about what is actually going on.

It would be great if there was documentation explaining precisely what the requirements for Rust types are so they can be placed into the eBPF map types provided by RedBPF.

It might also be a good idea to mark all methods exposed by the map types as `unsafe`. The reason is that normally, you can rely on Rust's type system to tell you whether your arguments have acceptable types. But here there are secondary requirements that cannot be expressed using trait bounds, but will nevertheless result in catastrophic program failure when disregarded. That's pretty much the definition of unsafety.

rhdxmr:
I wrote some note about valid `struct`s for BPF maps. I hope this note will be written to other forms of documentation later.


While developing BPF programs that deal with BPF maps, errors due to illegal
data definition frequently occur. It is basically FFI and is naturally a part
of unsafe Rust, so we have to be careful. If something goes wrong, infamous
defined behaviors will happen.

Here I am focusing on defining `struct`s. `union`s and `enum`s are off the
table here.

I couldn't come up with all possible ways to define valid structures. But users
can avoid pesky problems if they follow the conservative rules I suggest below.

1. Use `#[repr(C)]` representation

    Since the default representation does not guarantee anything about the
    layout, we can not control anything when BPF verifier complaints about the
    structure. Hence, let's use `repr(C)` always and control the layout of the
    structure and understand the specific size of the structure and figure out
    the alignments of all fields in the structure. Handling BPF maps is
    conducted by calling BPF helpers and it is naturally FFI. So `repr(C)`
    deserves to be used in BPF programs.

2. Add padding fields explicitly

    You may see the error if you don't add padding fields correctly:
[...]

PR #197, opened 2021-10-02
Support recovering original function parameters in retprobes

p-e-w:
### Background

Many functions in both kernel and user space take pointer arguments and manipulate them. An example is `tcp_v4_connect` from `net/ipv4/tcp_ipv4.c`, which implements most of the logic for the `connect` syscall. Its signature is

```
int tcp_v4_connect(struct sock *sk, struct sockaddr *uaddr, int addr_len)
```

The first parameter points to a struct that receives information about the connection, such as the destination address and port in the `skc_daddr` and `skc_dport` fields, respectively. When trying to use the kprobes mechanism to collect information about connections through `tcp_v4_connect`, this presents a problem:

* On function entry (kprobe), the struct `sk` has not been populated yet with the relevant information.
* On function exit (kretprobe), the struct has been populated, *but the RDI register (which holds the first parameter in the SysV calling convention) no longer points to the struct,* since the function code has almost certainly overwritten it.

Thus neither a kprobe nor a kretprobe can retrieve the required information *by itself.* The standard solution to this problem, [seen for instance in this `bcc` example](https://github.com/iovisor/bcc/blob/d3ceae0070470f0deec0bca5be51df4ebfe057d1/examples/tracing/tcpv4connect.py), is to register both a probe and a retprobe, and use a hash map, indexed by the current thread ID, to pass the parameter between them. Indeed, foniod itself [uses this approach in one of its probes](https://github.com/foniod/foniod/blob/df6bdd2ba617342a7befc6dccd5ed1f9306e7559/ingraind-probes/src/network/main.rs). Paraphrased from that probe, the relevant code looks like this:

```rust
#[map]
static mut SOCKETS: HashMap<u64, *const sock> = HashMap::with_max_entries(10240);

#[kprobe("tcp_v4_connect")]
fn tcp_v4_connect_enter(regs: Registers) {
    unsafe {
        SOCKETS.set(&bpf_get_current_pid_tgid(), &(regs.parm1() as *const sock));
    }
}

#[kretprobe("tcp_v4_connect")]
fn tcp_v4_connect_exit(regs: Registers) {
    let pid_tgid = bpf_get_current_pid_tgid();

    let socket = unsafe {
        match SOCKETS.get(&pid_tgid) {
            Some(s) => &**s,
            None => return,
        }
    };

    unsafe {
        SOCKETS.delete(&pid_tgid);
    }

    // Do something with `socket`
}
```

That's a lot of boilerplate for such a common task – and you need to repeat most of it for every instrumented function where the retprobe requires access to one of the original parameters!

### This PR

The purpose of this PR is to provide built-in support for the pattern described above. With the new feature, available on both the `kretprobe` and `uretprobe` macros, the equivalent code to the above snippet is

```rust
#[kretprobe]
fn tcp_v4_connect(regs: Registers, parms: [u64; 5]) {
    let socket = unsafe { &*(parms[0] as *const sock) };

    // Do something with `socket`
}
```

That's it! The macro detects that the annotated function takes a second argument, and **automatically** generates a probe/retprobe pair, a map holding parameter sets, and the code to populate/retrieve/clean up values from the map. The retprobe function is then called with the recovered original parameters. Note that if the second argument is not present in the signature of the annotated function, the macro generates exactly the same code as before, so **this change is fully backwards compatible.**

I have tested this feature with a variety of kprobe constellations. Please let me know if you have a specific scenario in mind that I should pay attention to.

### Why not save memory by sharing one parameter map between all retprobes?

Because this breaks when two functions are being probed, one of which calls the other. The thread ID will be the same in this case, and the inner function will first overwrite, then delete, the outer function's parameter set in the map. By the time the outer function's retprobe is invoked, there are no parameters for the thread and the handler will not be called.

### Why not store the full set of registers in the map, i.e. make the signature `retprobe(regs: Registers, orig_regs: Registers)`?

Because the memory layout of the underlying `pt_regs` struct prevents it from being stored in eBPF maps (see #150 for more information). This would also waste memory, since in virtually all cases the parameters are all that needs to be preserved but the full register context is substantially larger.

### The future: `fexit` probes

Kernel 5.5 introduced the new `fentry`/`fexit` BPF program types. In many ways, those represent a superior alternative to kprobes. One notable difference is that `fexit` probes have first-class access to the original function parameters without resorting to map storage or other workarounds. It would be great to see `fentry`/`fexit` support in RedBPF at some point, so that users of modern kernels can leverage those new features. But that is beyond the scope of this PR.

### DIFF
--- a/redbpf-macros/src/lib.rs
+++ b/redbpf-macros/src/lib.rs
@@ -309,6 +309,58 @@ fn probe_impl(ty: &str, attrs: TokenStream, item: ItemFn, mut name: String) -> T
     tokens.into()
 }
 
+fn probe_pair_impl(pre: &str, attrs: TokenStream, item: ItemFn, mut name: String) -> TokenStream {
+    if !attrs.is_empty() {
+        name = match parse_macro_input!(attrs as Expr) {
+            Expr::Lit(ExprLit {
+                lit: Lit::Str(s), ..
+            }) => s.value(),
+            _ => panic!("expected string literal"),
+        }
+    };
+
+    let ident = item.sig.ident.clone();
+    let map_ident = Ident::new(&format!("PARMS_{}", ident), Span::call_site());
+    let enter_ident = Ident::new(&format!("enter_{}", ident), Span::call_site());
+    let exit_ident = Ident::new(&format!("exit_{}", ident), Span::call_site());
+    let probe_ident = Ident::new(&format!("{}probe", pre), Span::call_site());
+    let retprobe_ident = Ident::new(&format!("{}retprobe", pre), Span::call_site());
+
+    let tokens = quote! {
+        #[map]
+        static mut #map_ident: HashMap<u64, [u64; 5]> = HashMap::with_max_entries(10240);
+
+        #[#probe_ident(#name)]
+        fn #enter_ident(regs: Registers) {
+            let pid_tgid = bpf_get_current_pid_tgid();
+            let parms = [regs.parm1(), regs.parm2(), regs.parm3(), regs.parm4(), regs.parm5()];
+            unsafe {
+                #map_ident.set(&pid_tgid, &parms);
+            }
+        }
+
+        #[#retprobe_ident(#name)]
+        fn #exit_ident(regs: Registers) {
+            let pid_tgid = bpf_get_current_pid_tgid();
+            let parms = unsafe {
+                match #map_ident.get(&pid_tgid) {
+                    Some(parms) => {
+                        let parms = *parms;
+                        #map_ident.delete(&pid_tgid);
+                        parms
+                    }
+                    None => return,
+                }
+            };
+            let _ = unsafe { #ident(regs, parms) };
+        }
+
+        #item
+    };
+
+    tokens.into()
+}
+
 fn wrap_kprobe(item: ItemFn) -> ItemFn {
     let ident = item.sig.ident.clone();
     let outer_ident = Ident::new(&format!("outer_{}", ident), Span::call_site());
@@ -345,6 +397,7 @@ pub fn kprobe(attrs: TokenStream, item: TokenStream) -> TokenStream {
 /// Attribute macro that must be used to define [`kretprobes`](https://www.kernel.org/doc/Documentation/kprobes.txt).
 ///
 /// # Example
+///
 /// ```no_run
 /// use redbpf_probes::kprobe::prelude::*;
 ///
@@ -353,10 +406,44 @@ pub fn kprobe(attrs: TokenStream, item: TokenStream) -> TokenStream {
 ///     // this is executed when clone() returns
 /// }
 /// ```
+///
+/// # Function parameters
+///
+/// In general, the `parmX` methods of the `regs` argument do **not**
+/// return the original parameter values that the function being probed
+/// was called with. The reason is that those parameters are passed
+/// via (architecture-dependent) general purpose registers, and the
+/// function code most likely overwrites some or all of those registers.
+/// RedBPF provides a convenient way to access original function parameters
+/// by declaring the retprobe with an additional array argument that
+/// receives function parameters 1-5:
+///
+/// ```no_run
+/// use redbpf_probes::kprobe::prelude::*;
+///
+/// #[kretprobe("__x64_sys_clone")]
+/// fn clone_exit(regs: Registers, parms: [u64; 5]) {
+///     // this is executed when clone() returns
+/// }
+/// ```
+///
+/// To make this possible, RedBPF generates a global map, and an entry probe
+/// corresponding to the retprobe which stores the original parameters
+/// in that map. A generated retprobe wrapper retrieves the parameters from
+/// the map, and calls the provided function with the parameter array as
+/// an argument.
+///
+/// Note that if no parameters for the current thread are found in the map
+/// (for example because the capacity of the map has been exhausted, or
+/// the retprobe was registered after the function had already been entered),
+/// **the retprobe is not called** for that function invocation.
 #[proc_macro_attribute]
 pub fn kretprobe(attrs: TokenStream, item: TokenStream) -> TokenStream {
     let item = parse_macro_input!(item as ItemFn);
     let name = item.sig.ident.to_string();
+    if item.sig.inputs.len() == 2 {
+        return probe_pair_impl("k", attrs, item, name);
+    }
     let wrapper = wrap_kprobe(item);
     probe_impl("kretprobe", attrs, wrapper, name)
 }
@@ -383,6 +470,7 @@ pub fn uprobe(attrs: TokenStream, item: TokenStream) -> TokenStream {
 /// Attribute macro that must be used to define [`uretprobes`](https://www.kernel.org/doc/Documentation/trace/uprobetracer.txt).
 ///
 /// # Example
+///
 /// ```no_run
 /// use redbpf_probes::uprobe::prelude::*;
 ///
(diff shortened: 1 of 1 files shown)


rhdxmr:
Hello @p-e-w

This PR shows me very clever idea. This approach is novel to me and I like it.
It also reduces boilerplate and makes code simple.

This PR introduces a new public API, so I checked whether API is clear and
stable. As you know it is difficult to change API once it is opened to
public. We will be able to change the API when we release redbpf v3.0 because we
should comply the semver semantic. So forgive me even if I seem pedantic.

At first glance, I didn't like the array. But shortly I was persuaded why you
chose the array instead of structure. And now it looks neat and intuitive to
use the array as original parameters.

But I wondered why the array size is 5? How about making it 6? I know that
x86-64 can pass 6 arguments via registers. So IMHO 6 looks more reasonable.

----
I realized that redbpf does not support to have multiple BPF programs in
one section. Phew..

So if I change the [`vfsreadlat
example`](https://github.com/foniod/redbpf/blob/main/examples/example-probes/src/vfsreadlat/main.rs)
to have the original parameters, then it is failed to load the program. Because
`probe_pair_impl` automatically defines a new function for `kprobe` with the
same name.

```rust
#[kretprobe("vfs_read")]
fn vfs_read_exit(regs: Registers, parms: [u64; 5])
```

```
ERROR redbpf: error loading BPF program `vfs_read' with bpf_load_program_xattr: unreachable insn 23
processed 0 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0

thread 'main' panicked at 'error loading BPF program: LoadError("vfs_read", BPF)', examples/example-userspace/examples/vfsreadlat.rs:95:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I think it's better to discuss this problem through another issue. Because this
problem has been in existence for a long time...

----

I wonder `PerCpuHashMap` is used instead of `HashMap`. I am confused whether
there are race conditions.

p-e-w:
Thank you for your feedback, @rhdxmr! I will try to address your points here:

> But I wondered why the array size is 5? How about making it 6? I know that x86-64 can pass 6 arguments via registers. So IMHO 6 looks more reasonable.

I chose to expose 5 parameters because that's what `Registers` does: It has methods `parm1` to `parm5`, but not `parm6`. Different calling conventions have different number of parameters that can be passed via registers; for example, ARM64 can pass 8, while ARM32 can only pass 4.

Whatever the "correct" number is here, I hope we agree that the array argument should expose the same number of parameters as the `Registers` struct.

> I realized that redbpf does not support to have multiple BPF programs in one section. Phew..
>
> [...]
>
> I think it's better to discuss this problem through another issue. Because this problem has been in existence for a long time...

I'm not sure if the problem is with RedBPF. In fact, I'm not even sure if there is a problem at all.

I believe the error message is actually coming from the kernel's eBPF bytecode verifier. When an eBPF program defines two probes for the same function, only one of them will be called, which makes the other one "unreachable code". The verifier in the kernel detects this, and refuses to load the program into the virtual machine.

I may be completely wrong here, but if that's what is going on, the only thing to fix would be the error message to make it clearer what the issue is. But this is true for most bugs in eBPF programs; the error message rarely gives even a hint about what's actually happening.

> I wonder `PerCpuHashMap` should be used instead of `HashMap`. I am confused whether there are race conditions. Do you know about this?

You are touching on a very difficult question here: Is all of this actually thread safe / safe from data races?

I have spent quite some time researching this topic, and the official documentation is incredibly vague. There is [an LWN article titled "Concurrency management in BPF"](https://lwn.net/Articles/779120/) which talks about unspecified "concurrency problems" with BPF maps, but the remaining paragraphs suggest that these apply only when map values are being modified *in place.*

The clearest statement I found comes from [the Android source code](https://android.googlesource.com/platform/system/bpf/+/75b410bdf186263fa4e05e079bfba44578622c33/libbpf/include/bpf/BpfMap.h#43):

```c
// Though the map class object and the underlying kernel map are thread safe, it
// is not safe to iterate over a map while another thread or process is deleting
// from it. In this case the iteration can return duplicate entries.
```

We don't iterate over the map used to store parameter sets, so this isn't a problem. Map `set`/`get`/`delete` operations for different keys should not interfere with each other, and since by construction, different threads never access the same key (because the key *is* the thread ID), there shouldn't be a problem. I wish I had something more authoritative to offer, but AFAICT the kernel docs *simply don't talk about this.* Perhaps the most reassuring fact is that the `HashMap` approach appears to be used by every other publicly available eBPF program that requires function parameters in kretprobes.

Last but not least, I don't believe switching to `PerCpuHashMap` would solve anything. AFAICT (and once again, kernel documentation is almost nonexistent for that matter), `PerCpuHashMap` is a performance optimization, not a safety mechanism. The point is to have a separate internal map for each CPU, so that certain atomic operations (such as incrementing counters) don't require expensive cross-CPU synchronization. But I haven't found anything suggesting that `PerCpuHashMap` is something like a thread safe version of `HashMap`.

rhdxmr:
@p-e-w Thank you for sharing your opinion and your knowledge.

> Different calling conventions have different number of parameters that can be passed via registers; for example, ARM64 can pass 8, while ARM32 can only pass 4.

It makes sense. Different architecture supplies the different number of
registers. So, I changed my opinion and I am retracting that 6 looks reasonable because x86-64
supplies 6 registers.

But also we can easily add the `parm6` method to `Registers` if we decide to do it.
So we don't need to be stuck to 5 registers.

To aid in finding a proper number I investigated the kernel source code to see the stats of the number of
function parameters.

The result is below:
```console
# bpftool btf dump file /sys/kernel/btf/vmlinux | awk '/FUNC_PROTO/{match($0, "vlen=([0-9]*)", groups); stats[groups[1]]++}END{for (vlen in stats) print vlen, stats[vlen]}'
0 111
1 5104
2 7919
3 5548
4 3120
5 1598
6 735
7 307
8 141
9 64
10 31
11 16
12 7
13 7
14 2
```

This result shows us there are 111 *0-parameters* functions, 5104 *1-parameter*
functions, ... and so on.

Thus, if we provide 5 registers than we can cover about 95% of all kernel
functions. So I agree that 5 is a good number.

----

I understand that the BPF map itself will not be a problem. The added code does
not iterate the map and it does not modify the value in place. Now I think that
we can merge this PR. But I hope @rsdy to review this PR and I'll be happy to hear
his opinion.

----

Actually I am still confused but we don't need to make this clear now and it is
not a reason to block this PR merged at all.

I'll say what I makes me confused below just for reference. I may be wrong because I
don't understand the internal of the Linux kernel perfectly..

1. There are multiple CPU cores in the machine so several kernel functions are
running simultaneously.

2. So I guess that the same kernel function that a BPF program is currently
attached to can run simultaneously by other CPU cores.

3. In this case is it possible that the `bpf_get_current_pid_tgid` BPF helper
function returns the same value?

I am just curious because I love Linux and I want to understand it more.

----

I approve this PR.
@rsdy Can you review this?

regards,

rhdxmr (approved):

rsdy (approved):

rsdy:
I'm happy with this. Initially I was a bit concerned about the background compatibility, but this looks like an easy way to extend this interface.

One thing is that the ergonomics could actually go one step further: Instead of leaving the user to dereference the register, we may want to provide a syntax (or something to this effect) that does the typesafe conversion like so:

```
#[kretprobe(track_params = [0])]
fn tcp_v4_connect(regs: Registers, socket: *const sock) {
    // Do something with `socket`
}
```

@rhdxmr re: your questions, relying on `bpf_get_current_pid_tgid` is safe because the kprobe hooks will run on the same cpu that the underlying function runs on. Threads are tracked by their tgid, which is recovered on context switching, so this should be safe even if the kernelspace is preemptible. One thread will never run simultaneously on multiple cores. I'm a bit hazy about the specifics here, though.

PR #207, opened 2021-10-25
expose result of map set and add insert, update method

ihciah:
I want to set value in userspace, but I don't know if the set is succeed. Maybe the max size limit exceed, or other cases.
We may expost the result to users.

And I found the flag is not exposed either, which can be done by adding insert and update method with `BPF_NOEXIST` and `BPF_EXIST` flag.


rsdy:
I'm closing this as stale. Please re-open if this is still relevant.

PR #211, opened 2021-11-01
Add a method to detach an XDP program.

eero-thia:
This allows one to detach an XDP program from an individual interface
without dropping the entire XDP struct and detaching from all other
interfaces.

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -879,6 +879,40 @@ impl XDP {
         }
     }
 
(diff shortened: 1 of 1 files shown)


rsdy (approved):
Looks clean! Thanks, I think we've missed this for the longest!

rhdxmr:
Thank you. You found one of a pair that we had lost for a long time.

eero-thia:
Consider it a good sign that people out there are making use of your great work. :)

PR #216, opened 2021-11-09
sk_lookup loading support

raftario:
This pull request is pretty simple and just adds support for loading `sk_lookup` programs to the `redbpf` crate. I'm opening it as a draft as

1. Support hasn't been added to the rest of the crates
~~2. `sk_lookup` requires a relatively recent kernel version (5.9) and merging this would probably break things for many people~~

Point 1 could be addressed by simply building upon this to add support everywhere else (I'll probably do it at some point but I don't personally need it at the moment), ~~while for point 2 it might be interesting to consider feature-gating different program types.~~

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -56,10 +56,10 @@ pub mod xdp;
 
 pub use bpf_sys::uname;
 use bpf_sys::{
-    bpf_attach_type_BPF_SK_SKB_STREAM_PARSER, bpf_attach_type_BPF_SK_SKB_STREAM_VERDICT,
-    bpf_attach_type_BPF_TRACE_ITER, bpf_create_map_attr, bpf_create_map_xattr, bpf_insn,
-    bpf_iter_create, bpf_link_create, bpf_load_program_xattr, bpf_map_def, bpf_map_info,
-    bpf_map_type_BPF_MAP_TYPE_ARRAY, bpf_map_type_BPF_MAP_TYPE_HASH,
+    bpf_attach_type_BPF_SK_LOOKUP, bpf_attach_type_BPF_SK_SKB_STREAM_PARSER,
+    bpf_attach_type_BPF_SK_SKB_STREAM_VERDICT, bpf_attach_type_BPF_TRACE_ITER, bpf_create_map_attr,
+    bpf_create_map_xattr, bpf_insn, bpf_iter_create, bpf_link_create, bpf_load_program_xattr,
+    bpf_map_def, bpf_map_info, bpf_map_type_BPF_MAP_TYPE_ARRAY, bpf_map_type_BPF_MAP_TYPE_HASH,
     bpf_map_type_BPF_MAP_TYPE_LRU_HASH, bpf_map_type_BPF_MAP_TYPE_LRU_PERCPU_HASH,
     bpf_map_type_BPF_MAP_TYPE_PERCPU_ARRAY, bpf_map_type_BPF_MAP_TYPE_PERCPU_HASH,
     bpf_map_type_BPF_MAP_TYPE_PERF_EVENT_ARRAY, bpf_prog_type, BPF_ANY,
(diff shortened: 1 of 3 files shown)


rhdxmr:
Hello @raftario 

This feature looks cool!

Thanks to this PR I got to know about sk lookup feature.
I just read this article: https://lwn.net/Articles/819618/
(I don't understand all yet)

IMO it's okay to include new kernel features in RedBPF. Even though sk lookup feature
exists in RedBPF, users who don't use this feature can continue to use RedBPF
with the older Linux kernel. Because nothing would be executed unless they call
the new functions.

If users have problems while they try calling the sk lookup functions with the
older kernel, users should handle their kernel version issue. But brief note
about the minimum kernel version for sk lookup would be helpful to users.


And you concerned that you implemented only the userspace part of sk
lookup. Don't worry about this. Somebody or you can freely implement the half
part for BPF programs later.

redbpf/src/lib.rs
```diff
@@ -2280,7 +2357,8 @@ impl<'a> SockMap<'a> {
         Ok(SockMap { base: map })
     }
 
-    pub fn set(&mut self, mut idx: u32, mut fd: RawFd) -> Result<()> {
+    pub fn set(&mut self, mut idx: u32, fd: RawFd) -> Result<()> {
+        let mut fd = fd as u64;
```
rhdxmr, line 2361: @raftario 
I'm not sure why you changed this.
Can you elaborate?

raftario:
I'm not particularly familiar with the library so I was probably confused. I just assumed that since `bpf-sys` uses bindgen, and `bpf.h` in libbpf includes `linux/bpf.h`, the bindings for sk_lookup types would not be generated on older kernels since they aren't present in the kernel headers, and `use`ing them from `redbpf` would fail to compile on those systems.

redbpf/src/lib.rs
raftario: That was a temporary fix of mine, I tested my changes with the code from [this talk](https://ebpf.io/summit-2020-slides/eBPF_Summit_2020-Lightning-Jakub_Sitnicki-Steering_connections_to_sockets_with_BPF_socke_lookup_hook.pdf) which uses `sizeof(__u64)` as the value size for its socket map, so passing it a pointer to a 32bit value caused errors. This can probably be removed given the value size should be `sizeof(int)`.

raftario:
Would you be interested in merging this without support in the other crates ?

rhdxmr:
@raftario 
Yes, I welcome your contribution.
API for BPF programs can be implemented later.
Before this PR get merged to main branch, we have to make sure userspace API clear.

I wonder how you used this API. Can you add some example as a doc comment in the SkLookup structure?

raftario:
I'll add some documentation and convert to a proper PR then !

redbpf/src/lib.rs
```diff
@@ -1075,6 +1142,47 @@ impl SocketFilter {
...
+
+        Ok(())
+    }
+}
+
+impl Drop for SkLookup {
+    fn drop(&mut self) {
+        if let Some((nfd, lfd)) = self.link.take() {
+            unsafe {
+                libc::close(lfd);
```
rhdxmr, line 1179: @raftario 
I am confused that `bpf_link_detach` is needed or not.
- Only `close` is needed?
- `bpf_link_detach` has nothing to do with cleanup?
- `bpf_link_detach` is called first and then `close` should be called?

Can you explain about that?

raftario: If you never pin the link, it'll automatically be cleaned up when the file descriptor is closed

rhdxmr: Oh, I got it. Thank you.

rhdxmr (approved):
This PR is great! 👍 
Code is very neat and clean. Doc comment is elaborate.
Added code looks harmonious.

Thank you for your contribution I am going to merge this happily 🙋

Issue #218, opened 2021-11-13
bpf_timer is defined multiple times in build/redbpf-probes-xxx/out/gen_bindings.rs file


Sherlock-Holo:
I use redbpf 2.1 and cargo-bpf(with llvm13, command-line features), then I build my old project, but it reports

> ```
> error[E0428]: the name `bpf_timer` is defined multiple times
>       --> /home/sherlock/git/learn_bpf/target/bpf/release/build/redbpf-probes-22900dca2d1af05a/out/gen_bindings.rs:240875:5
>        |
> 139094 |     pub struct bpf_timer {
>        |     -------------------- previous definition of the type `bpf_timer` here
> ...
> 240875 |     pub struct bpf_timer {
>        |     ^^^^^^^^^^^^^^^^^^^^ `bpf_timer` redefined here
>        |
>        = note: `bpf_timer` must be defined only once in the type namespace of this module
> 
> error[E0119]: conflicting implementations of trait `core::clone::Clone` for type `bindings::gen_bindings::bpf_timer`
>       --> /home/sherlock/git/learn_bpf/target/bpf/release/build/redbpf-probes-22900dca2d1af05a/out/gen_bindings.rs:240874:27
>        |
> 139093 |     #[derive(Debug, Copy, Clone)]
>        |                           ----- first implementation here
> ...
> 240874 |     #[derive(Debug, Copy, Clone)]
>        |                           ^^^^^ conflicting implementation for `bindings::gen_bindings::bpf_timer`
>        |
>        = note: this error originates in the derive macro `Clone` (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> error[E0119]: conflicting implementations of trait `core::marker::Copy` for type `bindings::gen_bindings::bpf_timer`
>       --> /home/sherlock/git/learn_bpf/target/bpf/release/build/redbpf-probes-22900dca2d1af05a/out/gen_bindings.rs:240874:21
>        |
> 139093 |     #[derive(Debug, Copy, Clone)]
>        |                     ---- first implementation here
> ...
> 240874 |     #[derive(Debug, Copy, Clone)]
>        |                     ^^^^ conflicting implementation for `bindings::gen_bindings::bpf_timer`
>        |
>        = note: this error originates in the derive macro `Copy` (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> error[E0119]: conflicting implementations of trait `core::fmt::Debug` for type `bindings::gen_bindings::bpf_timer`
[...]

PR #223, opened 2021-11-15
Provide struct bpf_timer forward declaration along with vmlinux.h

rhdxmr:
Resolve the problem of `struct bpf_timer redefinition` reported by issue #218


rsdy (approved):
Just out of interest, do you think we have coverage for actually testing this code in the current matrix? I just want to avoid a situation where this accidentally breaks something while fixing it for the newer kernels.

rhdxmr:
@rsdy 
I think I don't comprehend what you asked to me because of lack of my English skills.

This change only affect the procedure that makes use of vmlinux.
I tested the modified code with kernel 5.15 that started to include `struct bpf_timer` in the vmlinux for the first time. I also tested it with kernel 5.13 that does not have `struct bpf_timer` in the vmlinux. And the `ubuntu-2104-build-test` tests generating rust bindings with vmlinux of the Linux kernel v5.11.

This PR just moves `struct bpf_timer;` from `vmlinux_supplement.h` to right after the `vmlinux.h`. It is just a forward declaration and there is no problem even if there exists real definition of `struct bpf_timer` in the `vmlinux.h`.

So I think it's okay to merge this PR.

rsdy:
@rhdxmr Yeah, that was my question. Thanks.

Issue #204, opened 2021-10-23
bpf tc can't load by tc or cargo bpf load


Sherlock-Holo:
I'm using Archlinux

> Linux sherlockholo-pc 5.14.14-arch1-1 #1 SMP PREEMPT Wed, 20 Oct 2021 21:35:18 +0000 x86_64 GNU/Linux

I try this codes
```rust
#![no_std]
#![no_main]

use redbpf_probes::tc::prelude::*;
// use redbpf_probes::xdp::prelude::*;

program!(0xFFFFFFFE, "GPL");

#[tc_action]
fn test(skb: SkBuff) -> TcActionResult {
    Ok(TcAction::Ok)
}
```

and build with `cargo bpf build`

When I try to load it by `sudo tc filter add ingress bpf da obj target/bpf/programs/learn_bpf/learn_bpf.elf sec tc_action/test`

it reports

> Error fetching program/map!
> Unable to load program

also I try to use cargo-bpf, `sudo -E cargo bpf load -i lo target/bpf/programs/learn_bpf/learn_bpf.elf`, it just stuck without any output.

How should I do to load the bpf tc program?

d0u9:
Is the rust code for bpf are directly compiled by llvm? Something like:

```
llvm [options] -arch bpf -o bpf.rust
```

Or multiple steps are utilized to translate .rust to .xxx to .bpf?

d0u9:
I just found that article, and that is much complex than I thought before.

rhdxmr:
@Sherlock-Holo 

> @rhdxmr well... that's so hard to try to remove the `.text` section :(
> 
> I have try similar bpf tc codes, written by c, and compile by `clang -I ./headers -Wall -O3 -target bpf -c main.c -o main.o` (please ignore the -I ./headers, actually there's nothing there)
> 
> and load by tc success directly, no need to remove any section
> 
> is a plan to let redbpf can load by tc directly without removing some section?

I agreed with your opinion so I tried to remove `.text` section from ELF relocatable.
The problem has been solved by this PR #224
The fix will be released soon with a RedBPF `v2.2.0` release.
Hence, with the `v2.2.0` perhaps don't need to struggle with removing the `.text` section manually.

Thanks,

PR #224, opened 2021-11-15
Remove .text section

rhdxmr:
* Remove `redbpf-macros/src/mem.rs` not to include `memcpy`, `memset`, `memcmp` in the `.text` section
* Strip `.text` section from the relocatable elf file

I was unconvinced why the `memcmp`, `memcpy`, `memset`, `memmove`, `bcmp` are defined in the `redbpf-macros`. Also, to strip `.text` section it is better to not have those definitions in `.text` section.


rhdxmr:
I tried to remove `.text` section to solve the problem that is reported by this issue
* #204

It is evident that all new users struggle with the `.text` section when they try using tc action.

rsdy (approved):

rsdy:
Is it worth cutting a release with this? Seems like this is a great improvement for many people.

rhdxmr:
@rsdy 
Okay. I'll prepare release note.
And I want one more PR to get merged before release new package.
I'll send new PR in few hours.

PR #228, opened 2021-11-19
Add support for sk_lookup BPF programs

cbrewster:
This follows up on https://github.com/foniod/redbpf/pull/216 on the BPF side of things.

Currently this is a work in progress, but should expose what is needed to write some `sk_lookup` programs.


cbrewster:
I don't think I'll have the time to fix this up and land 😢 anyone is welcome to pick this up though

PR #241, opened 2021-12-02
Support BTF for tc command

### LINKED ISSUE #233: BTF section is miss but bpf_spin_lock need it
When use `bpf_spin_lock` or `bpf_spin_unlock` in tc BPF program and load by tc, the verifier will reports

> ```
> 35: (85) call bpf_spin_lock#93
> map '' has to have BTF in order to use bpf_spin_lock
> ```

rhdxmr:
Currently, the `.BTF` section is removed during compilation if the ELF
relocatable file includes tc action probes. Because the `.BTF` section contains
invalid type identifiers that are not permitted by BPF verifier. The BPF
verifier allows type identifiers to have characters that are only compatible
with C language, but the type identifiers generated by rustc are invalid from
the point of view of BPF verifier. And also `.BTF` section contains new BTF
types that are not supported by `tc` command. So it was neat to delete the
problematic `.BTF` section from ELF relocatable file if it contains tc_action
BPF programs.

But a few days ago @Sherlock-Holo reported that `tc` commands fails to load BPF
program if it tries to call `bpf_spin_lock` helper function. And the root cause
is the absence of `.BTF` section. Because `bpf_spin_lock` needs BTF records.

To solve this issue, make `cargo-bpf` substitute invalid characters of type
identifiers with underscore and also modify `cargo-bpf` to filter out BTF types
that are not supported by `tc` command.

This resolves #233.


rsdy (approved):
Wow. This is impressive. 👍

rhdxmr:
I pushed new commits to fix build failure and now all build tests pass.

rhdxmr:
I am merging this PR. Thank you @rsdy for approving this.
I hope this PR is helpful to someone.

PR #244, opened 2021-12-11
Build RedBPF on aarch64 as well as x86_64 docker containers

rhdxmr:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -42,9 +42,9 @@ jobs:
         bash -c 'export KERNEL_SOURCE=$(echo /lib/modules/*) \
           && echo KERNEL_SOURCE=$KERNEL_SOURCE \
           && cargo clean \
-          && cargo build \
-          && cargo build --bin cargo-bpf \
-          && cargo build --features=kernel5_8 --examples'
+          && cargo build --no-default-features --features=llvm13 \
+          && cargo build --no-default-features --features=llvm13,command-line --bin cargo-bpf \
+          && cargo build --no-default-features --features=llvm13,kernel5_8 --examples'
     -
       name: Build RedBPF with vmlinux on x86_64 Ubuntu 21.04
       run: |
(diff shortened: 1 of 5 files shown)


rhdxmr:
Oops.. I mistakenly sent PR to this repo directly. My intention was to send it to my forked repo and test the yaml file.
I'll do some more test and request a review then.

rhdxmr:
@rsdy 

It took quite long time to finish this PR but finally it's done.
Phew.. making build images is a time consuming job.

Here's the list of modification this PR brings.

- Modify build-test.yml to use new `ghcr.io/foniod/redbpf-build` docker images
- Add `linux/arm64` build environments
- Fix build failure on `linux/arm64` build environments

.github/workflows/build-test.yml
```diff
@@ -14,129 +14,578 @@ on:
...
   publish:
     runs-on: ubuntu-latest
     if: startsWith( github.ref, 'refs/tags/v')
     needs:
-    - fedora-34-build-test
-    - ubuntu-2104-build-test
-    - ubuntu-2004-build-test
-    - ubuntu-1804-build-test
-    - alpine-314-build-test
+    - build-on-x86_64-ubuntu-2104
```
rsdy, line 575: I think we should split this list into 2 targets, x64 and aarch64, so that they can run in parallel. Then we can also make any other higher-level target depend on these two.

rsdy (approved):
It's an interesting approach to use, it looks cool. However, I'd maybe try to introduce separate build targets for x64 and aarch64 to make failures a bit easier to follow.

rhdxmr:
@rsdy 
I am merging this.

But I could not understand your suggestion.
What do you mean separate build targets for x64 and aarch64?

rsdy:
Just to categorise the jobs somehow to be able to say "x64 works" and "aarch64 works", without all the details. I just peeked at Actions, and it looks fine, so not a big deal. Sorry if the terminology was off.

Issue #239, opened 2021-11-29
cargo-bpf build errors on 64-bit ARM host


cheeseprocedure:
Hello,

I'm running into the following issue when attempting to install `cargo-bpf` on a 64-bit ARM VM. I *think* the issue is unique to ARM, as I can't reproduce it in a 64-bit Intel guest:

```
# cargo install cargo-bpf --no-default-features --features=llvm13,command-line
# ...
error[E0432]: unresolved import `super::__va_list_tag`
  --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bpf-sys-2.2.0/src/type_gen.rs:22:5
   |
22 |     __va_list_tag, btf, btf__free, btf__get_nr_types, btf__name_by_offset, btf__parse_elf,
   |     ^^^^^^^^^^^^^ no `__va_list_tag` in the root

error[E0308]: mismatched types
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/bpf-sys-2.2.0/src/type_gen.rs:186:22
    |
186 |                 Some(vdprintf_wrapper),
    |                      ^^^^^^^^^^^^^^^^ expected array `[u64; 4]`, found *-ptr
    |
    = note: expected fn pointer `unsafe extern "C" fn(_, _, [u64; 4])`
                  found fn item `unsafe extern "C" fn(_, _, *mut [type error]) {vdprintf_wrapper}`

Some errors have detailed explanations: E0308, E0432.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `bpf-sys` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `cargo-bpf v2.2.0`, intermediate artifacts can be found at `/tmp/cargo-installheCcej`

Caused by:
  build failed
```

I believe I have all the necessary build pre-reqs in place. System/tooling versions:

```
root@ebpflab1:~# uname -a
Linux ebpflab1 5.10.0-9-arm64 #1 SMP Debian 5.10.70-1 (2021-09-30) aarch64 GNU/Linux

# LLVM installed from apt.llvm.org repos
root@ebpflab1:~# llvm-config-13 --version
13.0.1

root@ebpflab1:~# rustc -vV
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
host: aarch64-unknown-linux-gnu
[...]

PR #245, opened 2021-12-12
fix: change symbol __va_list_tag to __gnuc_va_list for aarch64

ihciah:
This PR is to solve #239.
There's a problem: When to use `__gnuc_va_list` and when to use `__va_list_tag`? I use `__gnuc_va_list` for aarch64 and arm. I didn't test other arch.

### DIFF
--- a/bpf-sys/src/type_gen.rs
+++ b/bpf-sys/src/type_gen.rs
@@ -19,8 +19,8 @@ system.
 */
 
 use super::{
-    __va_list_tag, btf, btf__free, btf__get_nr_types, btf__name_by_offset, btf__parse_elf,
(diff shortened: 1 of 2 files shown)


rhdxmr:
@ihciah 
DCO (Developer Certificate Origin) is not satisfied so could you fix your commit message to contain `Signed-off-by:` and re-push with -f?

And don't care about other build-test failures because they are not caused by your commit. Recently the docker images were re-built but they are not configured properly. It will be fixed soon.

rhdxmr (approved):
This PR looks good to me.
It is intuitive understandable modification and it is proper solution.
Thanks for your contribution!

bobrik:
Could you publish a release with these changes?

rhdxmr:
@bobrik 
Thanks for asking me this.
Sure. I'll prepare new release within few days.

PR #246, opened 2021-12-16
Add support for devmap map type

kriomant:


rhdxmr:
@kriomant 
Apologies for the late response.

I reviewed this PR few times and checked it whether it works correctly. By writing some example code and running it in VM, I observed the DevMap is working as expected.

I think this PR is great and good to merge into main branch. Please don't mind the build test failure, I believe the failure is not caused by this PR. The build test had been fixed a few days after this PR was sent but as this PR is still based on the older commits, the build failure still happens.

Thanks for your contribution!
I was glad to look into your contribution and also I could learn from you. Thank you.

Happy New Year and I wish you all the best for 2022.

rhdxmr (approved):

rhdxmr:
@kriomant
You're welcome. It was my pleasure to review your PR.

> Can we always use "___btf_map{}" name for both cases? If not, why? 

tc or libbpf has its own format for delivering BTF information from probes to
runtime loader.  RedBPF also has its own semantic for doing that. And these
formats are internal implementation details.

Making RedBPF compatible with libbpf or tc is not a main goal of RedBPF project
at present so I decided to keep the tc-case separated from the normal case.

And the format of the BTF structure could be changed in the future. libbpf
could change the format without discussing with us, vice versa. So I think
it is better not to follow their internal implementation.

`____btf_map_{}` is the format of the name, that tc or libbpf make use of, so
RedBPF just mimics that. But as you know it is not a guaranteed thing like API,
but just an implementation detail of them. It is fragile.


> Can we also generate such struct for maps with fixed key and value types?

I'm sorry I don't understand what "fixed keys and values" means here.

kriomant:
> And the format of the BTF structure could be changed in the future. libbpf could change the format without discussing with us, vice versa. So I think it is better not to follow their internal implementation.
> `____btf_map_{}` is the format of the name, that tc or libbpf make use of, so RedBPF just mimics that. But as you know it is not a guaranteed thing like API, but just an implementation detail of them. It is fragile.

But right now you depend on it anyway, will it be any worse if you generate such info for all types of maps?

> > Can we also generate such struct for maps with fixed key and value types?
> I'm sorry I don't understand what "fixed keys and values" means here.

I mean maps which are not parametrised by key/value types, such as DevMap. They still has some internal key and value types, `u32` in case of DevMap.

If `map` macro generated same code for all types of maps (tc-compatible, parametrized by key/value types, non-parameterized) then it would have been possible to use traits. Benefits:
* user may use type aliases for maps, right now type name is analyzed
* user may define their own types of BPF maps compatible with `map` macro without modifying RedBPF library

rhdxmr:
@kriomant 
I understand what you are planning. That sounds cool.

> > And the format of the BTF structure could be changed in the future. libbpf could change the format without discussing with us, vice versa. So I think it is better not to follow their internal implementation.
> > `____btf_map_{}` is the format of the name, that tc or libbpf make use of, so RedBPF just mimics that. But as you know it is not a guaranteed thing like API, but just an implementation detail of them. It is fragile.
> 
> But right now you depend on it anyway, will it be any worse if you generate such info for all types of maps?

You're right. Currently, `____btf_map_{}` and `MapBtf` are similar so it will
not be any worse if `____btf_map_{}` type replaces `MapBtf` type.
I just looked into the RedBPF code related to the `MapBtf`, it seems okay to
unify the BTF type info. I'm not 100% sure so you might check that while you
are working on.


> If `map` macro generated same code for all types of maps (tc-compatible, parametrized by key/value types, non-parameterized) then it would have been possible to use traits. Benefits:
> 
> * user may use type aliases for maps, right now type name is analyzed
> * user may define their own types of BPF maps compatible with `map` macro without modifying RedBPF library

I like what you suggested as benefits. But I can't come up with how new API
would look like and how to make it better so I'd rather wait for your design
suggestion.


I appreciate your contribution to RedBPF.

Regards,

PR #250, opened 2022-01-02
Find if get or get_mut method of RedBPF maps are called

rhdxmr:
Previously cargo-bpf raises an error if the alignment of value is
greater than 8 bytes. But this is inappropriate to some BPF program that
never calls get/get_mut. Since calling these methods creates references
of possibly misaligned data, it is okay to permit values that have
alignment greater than 8 bytes if get/get_mut is not called anywhere in
BPF programs.

Find if get/get_mut method is called by inspecting LLVM bitcode and
raise an error.


rsdy (approved):
Definitely a huge improvement. Thanks!

PR #251, opened 2022-01-06
Add BPF_MAP_TYPE_RINGBUF support with RingBufMap

### LINKED ISSUE #161: Is there a plan for supporting BPF_MAP_TYPE_RINGBUF?
I have heard that the performance is better than BPF_MAP_TYPE_PERF_EVENT_ARRAY.

yobiscus:
Closes #161.

### DIFF
--- a/redbpf-probes/src/ringbuf.rs
+++ b/redbpf-probes/src/ringbuf.rs
@@ -1,0 +1,205 @@
+// Copyright 2022 Authors of Red Sift
+//
+// Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
+// http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
+// http://opensource.org/licenses/MIT>, at your option. This file may not be
+// copied, modified, or distributed except according to those terms.
+
+/*!
+eBPF ringbuf.
+ */
+use core::marker::PhantomData;
+use core::mem;
+use cty::*;
+
+use crate::bindings::*;
+use crate::helpers::*;
+use crate::maps::BpfMap;
+
+/// Flags that be passed to `RingBuf::output_with_flags`.
+#[derive(Debug, Copy, Clone)]
+pub struct RingBufMapFlags(u64);
+
+impl Default for RingBufMapFlags {
+    #[inline]
+    fn default() -> Self {
+        RingBufMapFlags(0)
+    }
+}
+
+impl RingBufMapFlags {
+    /// Create new default flags.
+    #[inline]
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Set BPF_RB_NO_WAKEUP flag.
+    #[inline]
+    pub fn no_wakeup(&mut self) -> &mut RingBufMapFlags {
+        self.0 |= BPF_RB_NO_WAKEUP as u64;
+        self
+    }
+
+    /// Set BPF_RB_FORCE_WAKEUP flag
+    #[inline]
+    pub fn force_wakeup(&mut self) -> &mut RingBufMapFlags {
+        self.0 |= BPF_RB_FORCE_WAKEUP as u64;
+        self
+    }
+}
+
+impl From<RingBufMapFlags> for u64 {
+    #[inline]
+    fn from(flags: RingBufMapFlags) -> u64 {
+        flags.0
+    }
+}
+
+/// Ring buffer map.
+///
+/// This is a wrapper for `BPF_MAP_TYPE_RINGBUF`.
+pub struct RingBufMap<T> {
+    def: bpf_map_def,
+    _event: PhantomData<T>,
+}
+
+impl<T> RingBufMap<T> {
+    /// Creates a ring buffer map with the specied maximum number of elements.
+    ///
+    /// `buffer_size` must be a power of 2 value and must be page aligned.
+    pub const fn with_buffer_size(buffer_size: u32) -> Self {
+        Self {
+            def: bpf_map_def {
+                type_: bpf_map_type_BPF_MAP_TYPE_RINGBUF,
+                key_size: 0,
+                value_size: 0,
+                max_entries: buffer_size,
+                map_flags: 0,
+            },
+            _event: PhantomData,
+        }
+    }
+
+    /// Copy data to ring buffer.
+    ///
+    /// The ring buffer can hold up to `buffer_size` bytes, see `with_buffer_size`.
+    /// No loss reporting occurs in user space. Monitoring must be done by the BPF
+    /// program.
+    ///
+    /// This method will wake up the polling thread if it is currently waiting on the
+    /// ring buffer. If this isn't desired, see `output_with_flags`.
+    #[inline]
+    pub fn output(&mut self, data: &T) -> Result<(), i64> {
+        self.output_with_flags(data, RingBufMapFlags::default())
+    }
+
+    /// Copy data to ring buffer specifying wake options in the given
+    /// `RingBufMapFlags`.
+    #[inline]
+    pub fn output_with_flags(&mut self, data: &T, flags: RingBufMapFlags) -> Result<(), i64> {
+        unsafe {
+            let ret = bpf_ringbuf_output(
+                &mut self.def as *mut _ as *mut c_void,
+                data as *const _ as *mut c_void,
+                mem::size_of::<T>() as u64,
+                flags.into(),
+            );
+            if ret == 0 {
+                Ok(())
+            } else {
+                Err(ret)
+            }
+        }
+    }
+
+    /// Reserve sample in ring buffer.
+    #[inline]
+    pub fn reserve(&mut self) -> Option<RingBufPayload<T>> {
+        unsafe {
+            let ptr = bpf_ringbuf_reserve(
+                &mut self.def as *mut _ as *mut c_void,
+                mem::size_of::<T>() as u64,
+                0,
+            ) as *mut T;
+
+            if ptr.is_null() {
+                None
+            } else {
+                Some(RingBufPayload {
+                    reserved: true,
+                    ptr,
+                })
+            }
+        }
+    }
+}
+
+impl<T> BpfMap for RingBufMap<T> {
+    type Key = u32;
+    type Value = T;
+}
+
+/// Ring buffer map entry.
+pub struct RingBufPayload<T> {
+    reserved: bool,
+    ptr: *mut T,
+}
+
+impl<T> RingBufPayload<T> {
+    /// Retrieve payload as mutable reference.
+    ///
+    /// # Safety
+    ///
+    /// Safety of this funciton is the same as ptr::as_mut:
+    ///
+    /// * The pointer must be properly aligned.
+    ///
+    /// * It must be "dereferencable" in the sense defined in [the module documentation].
+    ///
+    /// * You must enforce Rust's aliasing rules, since the returned lifetime `'a` is
+    ///   arbitrarily chosen and does not necessarily reflect the actual lifetime of the data.
+    ///   In particular, for the duration of this lifetime, the memory the pointer points to must
+    ///   not get accessed (read or written) through any other pointer.
+    ///
+    /// This pointer is already assumed to be non-null since it was checked in RingBufMap::reserve.
+    pub unsafe fn as_mut(&self) -> &mut T {
+        self.ptr.as_mut().unwrap()
+    }
+
+    /// Submit reserved sample.
+    #[inline]
+    pub fn submit(&mut self) {
+        self.submit_with_flags(RingBufMapFlags::default())
+    }
+
+    /// Submit reserved sample specifying wake options in the given `RingBufMapFlags`.
+    #[inline]
+    pub fn submit_with_flags(&mut self, flags: RingBufMapFlags) {
+        unsafe { bpf_ringbuf_submit(self.ptr as *mut c_void, flags.into()) }
+        self.reserved = false;
+    }
+
+    /// Discard reserved sample.
+    #[inline]
+    pub fn discard(&mut self) {
+        self.discard_with_flags(RingBufMapFlags::default());
+    }
+
+    /// Discard reserved sample specifying wake options in the given `RingBufMapFlags`.
+    #[inline]
+    pub fn discard_with_flags(&mut self, flags: RingBufMapFlags) {
+        unsafe {
+            bpf_ringbuf_discard(self.ptr as *mut c_void, flags.into());
+        }
+        self.reserved = false;
+    }
+}
+
+impl<T> Drop for RingBufPayload<T> {
+    fn drop(&mut self) {
+        if self.reserved {
+            self.discard();
+        }
+    }
+}
--- a/redbpf/src/load/loader.rs
+++ b/redbpf/src/load/loader.rs
@@ -12,11 +12,11 @@ use std::fs;
 use std::io;
 use std::path::Path;
 
-use crate::load::map_io::PerfMessageStream;
+use crate::load::map_io::{PerfMessageStream, RingBufMessageStream};
 use crate::{cpus, Program};
 use crate::{
-    Error, KProbe, Map, Module, PerfMap, SkLookup, SocketFilter, StreamParser, StreamVerdict,
-    TaskIter, UProbe, XDP,
+    Error, KProbe, Map, Module, PerfMap, RingBufMap, SkLookup, SocketFilter, StreamParser,
+    StreamVerdict, TaskIter, UProbe, XDP,
 };
 
 #[derive(Debug)]
@@ -44,6 +44,7 @@ impl Loader {
 
         let online_cpus = cpus::get_online().unwrap();
         let (sender, receiver) = mpsc::unbounded();
+
         // bpf_map_type_BPF_MAP_TYPE_PERF_EVENT_ARRAY = 4
         for m in module.maps.iter_mut().filter(|m| m.kind == 4) {
             for cpuid in online_cpus.iter() {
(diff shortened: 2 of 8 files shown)


yobiscus:
RINGBUF is a relatively new addition to the kernel. Should I put it behind a feature gate (default disabled)?

rhdxmr:
Hello @yobiscus 

It looks great! I feel interested in this PR a lot. Thanks for your
contribution :)
Actually I was working on ringbuf as well so I can discuss ringbuf.
I am glad to see this PR since it is a good opportunity to talk about the same
topic we both are interested in.

But first, it would be better to fix the build failure issue. I also feel embarassed
to see this failure since I also didn't experience this failure in my latest
development environment (fedora 35). But that's the point of distro build
tests. RedBPF tries to support the latest Ubuntu LTS which is 20.04 now.

Currently the default kernel of ubuntu 20.04 LTS is v5.4 which does not support
ring buffer. It's a General Availability(GA) kernel of ubuntu. So it would be
nice to make ringbuf feature as an optional feature not to cause build failure
in ubuntu 20.04 LTS. And nevermind ubuntu 18.04.. it's older ubuntu LTS so I'm
going to disable it soon.

yobiscus:
Hey @rhdxmr, thanks for your interest. Personally I have been testing on Centos 8.5, which uses a slightly more awkward versioning system for the kernel (kernel version is 4.18 officially, but in reality it has all the BPF features of kernel 5.9).

I'll spin up a 20.04 VM and fix the build here with conditional compilation. It might take me a few days to learn about conditional compilation and to juggle other priorities, so thanks in advance for your patience :)

rhdxmr:
Hey, @yobiscus 
> Personally I have been testing on Centos 8.5, which uses a slightly more
  awkward versioning system for the kernel (kernel version is 4.18 officially,
  but in reality it has all the BPF features of kernel 5.9).

4.18 + BPF of 5.9?! That sounds peculiar to me. But I'm just curious about why
they did that. Anyway it is interesting to see that they think BPF is important.

> It might take me a few days to learn about conditional compilation and to
  juggle other priorities

It's absolutely fine. Thank you for your effort.

yobiscus:
I fixed the build in Ubuntu 20.04 and confirmed that use of ringbuf still works when the feature is explicitly enabled (on systems that support it). I realize documentation and examples may be useful to add as well. I can add those once we're happy with the implementation.

redbpf/src/ringbuf.rs
```diff
@@ -0,0 +1,110 @@
...
+                    ring_buffer: RingBufInner(ring_buffer),
+                    fd,
+                })
+            }
+        }
+    }
+
+    pub fn read_messages(&self) -> Vec<Box<[u8]>> {
+        unsafe {
+            ring_buffer__consume(self.ring_buffer.0);
```
rhdxmr, line 98: Wow this is great :+1:
Although `ring_buffer_sample_cb` stored previous messages into buffer, we could consume some new records from the ring buffer, that are not yet notified.

redbpf/src/ringbuf.rs
```diff
@@ -0,0 +1,110 @@
...
+        let mut messages = self.messages.borrow_mut();
+        let new_vec = messages.clone();
+        messages.clear();
+        new_vec
+    }
+}
+
+#[derive(Debug)]
+pub struct RingBufInner(*mut ring_buffer);
+unsafe impl Send for RingBufInner {}
```
rhdxmr, line 57: I wonder this is required by `tokio`?

redbpf/src/load/loader.rs
```diff
@@ -59,6 +60,19 @@ impl Loader {
...
+            let name = m.name.clone();
+            let map = RingBufMap::bind(m).unwrap();
+            let stream = RingBufMessageStream::new(name.clone(), map);
+            let mut s = sender.clone();
+            let fut = stream.for_each(move |events| {
+                s.start_send((name.clone(), events)).unwrap();
+                future::ready(())
+            });
+            tokio::spawn(fut);
+        }
```
rhdxmr, line 74: Wow. This is nice to offer users the same interface compatible with perfmap.

redbpf/src/ringbuf.rs
```diff
@@ -0,0 +1,110 @@
...
+
+unsafe extern "C" fn ring_buffer_sample_cb(ctx: *mut c_void, data: *mut c_void, size: u64) -> i32 {
+    let rb = &mut *(ctx as *mut RingBufMapContext);
+    rb.read_message(data, size);
+    0
+}
+
+impl RingBufMap {
+    pub fn bind(map: &mut Map) -> Result<RingBufMap> {
+        unsafe {
```
rhdxmr, line 74: It might be better to check some fields of `map.config`. To prevent incompatible map from being passed to this function by mistake.

redbpf/src/load/map_io.rs
```diff
@@ -69,3 +69,38 @@ impl Stream for PerfMessageStream {
         Some(self.read_messages()).into()
     }
 }
+
+pub struct RingBufMessageStream {
```
rhdxmr, line 73: Is it necessary to export this to public?
Is it technical requirement to implement this feature or do you mean this to be used by users directly?

redbpf/src/ringbuf.rs
```diff
@@ -0,0 +1,110 @@
...
+use std::os::unix::io::RawFd;
+use std::slice;
+
+use bpf_sys::{
+    ring_buffer, ring_buffer__consume, ring_buffer__epoll_fd, ring_buffer__free, ring_buffer__new,
+};
+use libc::c_void;
+
+#[derive(Debug)]
+pub struct RingBufMapContext {
```
rhdxmr, line 28: Do you expect this API will be used by users directly?

rhdxmr (changes requested):

rhdxmr:
Hello @yobiscus

I checked your code carefully and I leaved review messages.
Some of them are my opinion and perhaps seem pedantic
so if you have better idea, please share your thought with me.

I hope this review helpful.

redbpf-probes/src/maps.rs
```diff
@@ -392,6 +392,150 @@ impl<T> PerfMap<T> {
...
+    pub struct RingBufMap<T> {
+        def: bpf_map_def,
+        _event: PhantomData<T>,
+    }
+
+    impl<T> RingBufMap<T> {
+        /// Creates a ring buffer map with the specied maximum number of elements.
+        ///
+        /// `max_entries` must be a power of 2 value and must be page aligned.
+        pub const fn with_max_entries(max_entries: u32) -> Self {
```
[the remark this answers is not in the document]
rsdy: `max_entries` keeps it consistent with other maps of the like, so I think it's preferable.

redbpf-probes/src/maps.rs
```diff
@@ -392,6 +392,150 @@ impl<T> PerfMap<T> {
...
+        #[inline]
+        pub fn submit(&mut self, data: &mut T) {
+            self.submit_with_flags(data, RingBufMapFlags::default())
+        }
+
+        /// Submit reserved sample specifying wake options in the given `RingBufMapFlags`.
+        #[inline]
+        pub fn submit_with_flags(&mut self, data: &mut T, flags: RingBufMapFlags) {
+            unsafe { bpf_ringbuf_submit(data as *mut _ as *mut c_void, flags.into()) }
+        }
```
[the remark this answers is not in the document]
rsdy: This auxilliary struct could also help contain the received pointer. I think we could expose the ptr -> ref conversion through an `unsafe fn`, or through a `Result<&mut T, E>` gate, where the accessor function validates pointer alignment before creating the reference.

redbpf/src/load/map_io.rs
rsdy: A ring buffer is handy to be consumed through a Stream, so I think it's worth making this public. See the PerfMessageStream above, which was designed specifically around this usecase.

redbpf-probes/src/maps.rs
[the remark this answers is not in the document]
yobiscus: These are great suggestions. Using auxiliary struct with the Drop trait instead of an explicit discard should make probe logic cleaner and simpler.

redbpf-probes/src/maps.rs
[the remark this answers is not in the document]
yobiscus: I chose `max_entries` for consistency, but really this is allocating a buffer of a certain size rather than an array of elements. `with_buffer_size` is more descriptive.

redbpf/src/ringbuf.rs
yobiscus: Sorry, I'm not following ☺️. `ring_buffer__consume` is directly responsible for calling `ring_buffer_sample_cb`. `ring_buffer_sample_cb` is not invoked otherwise.

redbpf/src/ringbuf.rs
yobiscus: This should not be public. Thanks!

yobiscus:
@rsdy no problem, apologies for the delay. I was preoccupied with other duties over the last couple of weekends. Let me try to address all outstanding issues this coming weekend.

redbpf/src/ringbuf.rs
yobiscus: Yes, that's right:
```
error[E0277]: `*mut ring_buffer` cannot be sent between threads safely
   --> /home/vagrant/redbpf/redbpf/src/load/loader.rs:73:13
    |
73  |             tokio::spawn(fut);
    |             ^^^^^^^^^^^^ `*mut ring_buffer` cannot be sent between threads safely
    |
   ::: /home/vagrant/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.15.0/src/task/spawn.rs:127:21
    |
127 |         T: Future + Send + 'static,
    |                     ---- required by this bound in `tokio::spawn`
    |
    = help: within `ForEach<RingBufMessageStream, futures::future::Ready<()>, [closure@/home/vagrant/redbpf/redbpf/src/load/loader.rs:69:39: 72:14]>`, the trait `std::marker::Send` is not implemented for `*mut ring_buffer`
note: required because it appears within the type `RingBufInner`
   --> /home/vagrant/redbpf/redbpf/src/ringbuf.rs:56:12
    |
56  | pub struct RingBufInner(*mut ring_buffer);
    |            ^^^^^^^^^^^^
note: required because it appears within the type `ringbuf::RingBufMap`
   --> /home/vagrant/redbpf/redbpf/src/ringbuf.rs:59:12
    |
59  | pub struct RingBufMap {
    |            ^^^^^^^^^^
note: required because it appears within the type `RingBufMessageStream`
   --> /home/vagrant/redbpf/redbpf/src/load/map_io.rs:73:12
    |
73  | pub struct RingBufMessageStream {
    |            ^^^^^^^^^^^^^^^^^^^^
    = note: required because it appears within the type `ForEach<RingBufMessageStream, futures::future::Ready<()>, [closure@/home/vagrant/redbpf/redbpf/src/load/loader.rs:69:39: 72:14]>`

For more information about this error, try `rustc --explain E0277`.
error: could not compile `redbpf` due to previous error
error: failed to compile the `pidtree' program
```

redbpf-probes/src/maps.rs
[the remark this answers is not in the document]
yobiscus: Changed the name and docs. I think the docs better match the behaviour now.

redbpf-probes/src/maps.rs
```diff
@@ -392,6 +392,150 @@ impl<T> PerfMap<T> {
     }
 }
 
+#[cfg(feature = "ringbuf")]
+mod ringbuf {
```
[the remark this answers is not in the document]
yobiscus: Moved.

redbpf/src/ringbuf.rs
yobiscus: Is there any action item here for me to do?

redbpf/src/ringbuf.rs
yobiscus: Checking for map.kind now. Anything else?

redbpf/src/ringbuf.rs
yobiscus: Fixed.

redbpf-probes/src/maps.rs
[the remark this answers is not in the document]
yobiscus: Still outstanding. How do I go about validating pointer alignment here?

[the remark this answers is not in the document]
rhdxmr: How about passing value and write it to the dest using `ptr::write_unaligned`?

redbpf/src/ringbuf.rs
rhdxmr: I think this code looks okay :slightly_smiling_face:

redbpf/src/ringbuf.rs
rhdxmr: I think checking some fields of map.config will be enough :slightly_smiling_face:

redbpf-probes/src/maps.rs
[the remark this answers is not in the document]
yobiscus: Doesn't this defeat the optimizations of reserve/submit? The idea is to avoid creating temporary structures and avoiding copies altogether.

Can we make this the responsibility of the user of the API instead?

I created the auxiliary type as discussed. The user can chose to use the unsafe `as_mut` helper, or access the ptr directly from the struct. What do you think?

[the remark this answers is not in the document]
yobiscus: For reference, this is how I'm using the API now. I think it looks better now!

```
+        let mut payload = NEW_TASKS.reserve().ok_or(ProbeError::RingBufReserveError)?;
+        let event = payload.as_mut();
+        event.pid = task.0.pid as u64;
+        event.ppid = bpf_get_current_pid_tgid() >> 32;
+        payload.submit();
```

yobiscus:
AFAIK all review comments. Please let me know if there are other outstanding items. Thanks!

rhdxmr (commented):
Thanks for your follow-up modifications.

I added some comment.

And please modify your new 3 commit messages to contain `Signed-off-by` trailer.

redbpf/src/load/loader.rs
```diff
@@ -59,6 +60,19 @@ impl Loader {
...
+        // bpf_map_type_BPF_MAP_TYPE_RINGBUF
+        for m in module.maps.iter_mut().filter(|m| m.kind == 27) {
+            let name = m.name.clone();
+            let map = RingBufMap::bind(m).unwrap();
+            let stream = RingBufMessageStream::new(name.clone(), map);
+            let mut s = sender.clone();
+            let fut = stream.for_each(move |events| {
+                s.start_send((name.clone(), events)).unwrap();
+                future::ready(())
+            });
```
nbaksalyar, line 72: I've used this branch to write some production-ish code and I have some feedback.

This works like a charm but I think the API here is a bit restrictive because there's no easy way to listen on ringbufs individually. Example case: I want to start my own async task for each ringbuf to forward incoming events to a TCP connection. With this approach, I'll have to filter all events to look for particular ringbufs that interest me, then likely forward these events through an mpsc queue, etc. etc.

Instead, I propose to store ringbufs in a separate `HashMap`:

```suggestion
            let mut ringbufs = HashMap::new();
            ...
            ringbufs.insert(name, stream);
```
and make it publicly accessible from `struct Loaded`, e.g.:

```rust
pub struct Loaded {
    pub events: mpsc::UnboundedReceiver<(String, <PerfMessageStream as Stream>::Item)>,
    //...
    pub ringbufs: HashMap<String, RingBufMessageStream>,
}
```

This would also simplify the implementation.

What do you think? cc @rhdxmr 

@yobiscus and thanks for working on this, it's an awesome feature! :)

rsdy: @nbaksalyar that makes a lot of sense.

I'm afraid this PR is a bit stale, so at this stage we might need another revision, + we can't merge without DCO from @yobiscus . Hang in there for another week or so to see if they re-appear, and we can figure out a way forward if not.

yobiscus: Thanks for the much needed poke @nbaksalyar :) I am currently updating my environment so that it is compatible with the latest redbpf changes. I have rebased my changes against the main branch and hope to push something functional tomorrow.

Unfortunately I have not found time to address @rhdxmr comments above, specifically the comment about pointer alignment. There's a bit of a learning curve needed to address this which I don't currently have bandwidth for. Any help closing on these issues after I rebase my branch is very much appreciated!

redbpf-probes/src/ringbuf.rs
```diff
@@ -0,0 +1,194 @@
...
+use core::marker::PhantomData;
+use core::mem;
+use cty::*;
+
+use crate::bindings::*;
+use crate::helpers::*;
+
+/// Flags that be passed to `RingBuf::output_with_flags`.
+#[derive(Debug, Copy, Clone)]
+pub struct RingBufMapFlags(u32);
```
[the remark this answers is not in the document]
yobiscus: Fixing this, thanks.

redbpf-probes/src/ringbuf.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// ring buffer. If this isn't desired, see `output_with_flags`.
+    #[inline]
+    pub fn output(&mut self, data: &T) -> i64 {
+        self.output_with_flags(data, RingBufMapFlags::default())
+    }
+
+    /// Copy data to ring buffer specifying wake options in the given
+    /// `RingBufMapFlags`.
+    #[inline]
+    pub fn output_with_flags(&mut self, data: &T, flags: RingBufMapFlags) -> i64 {
```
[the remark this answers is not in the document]
yobiscus: Thanks, I'll return Result<(), i64> instead.

redbpf/src/load/loader.rs
yobiscus: Pushed and signed-off. Sorry I wasn't able to implement your suggestion @nbaksalyar. Since this content is feature gated, would it be okay to merge this as is and address the comments after-the-fact, even if the changes are not backwards compatible?

nbaksalyar: Sure, no worries! I can help with making this change if you'd like.

rsdy (approved):

redbpf/src/load/loader.rs
rsdy: Thanks everyone, this looks good, once the CI passes I can merge this!

Issue #189, opened 2021-09-24
Linking failed with cdylib crate


sebastiaoamaro:
Hello everyone, I am trying to create a cdylib crate to be used as a python library using (pyo3) that has a function that reads from a pinned map.
 
However, I cannot build giving this error:

note: /"path"/target/release/deps/libbpf_sys-5ebb6e925d9bf342.rlib(libbpf.o): warning: relocation against `stderr@@GLIBC_2.2.5' in read-only section `.text'
          /"path"/target/release/deps/libbpf_sys-5ebb6e925d9bf342.rlib(libbpf.o): relocation R_X86_64_PC32 against symbol `stderr@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: final link failed: bad value
          collect2: error: ld returned 1 exit status

So I created a .cargo folder with a config file in it with the following line:

rustflags = ["-C", "link-args=-fPIC"]

This seems to add the flag:

"/home/sebas/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-o" "/home/sebas/tese/kollaps-private/communication/target/release/deps/libcommunication.so" "-Wl,--gc-sections" "-shared" "-Wl,-zrelro" "-Wl,-znow" "-Wl,-O1" "-nodefaultlibs" "-fPIC"

But still, the same error appears does anyone know why this might be happening and how to solve it?

Thanks in advance.

rsdy:
@sebastiaoamaro Is there a way you could publish a minimum reproducible example for this? Also, please take a look if the `main` branch solves your issue, as #183 could be a show stopper here.

sebastiaoamaro:
Hello, I solved the problem by adding the -fPIC flag to this file https://github.com/foniod/redbpf/blob/main/bpf-sys/build.rs (line 51) and using the dependency locally.

rhdxmr:
Wow @sebastiaoamaro
You made it! 
It guess that you are make something great with cutting edge technologies. I have never used `pyo3` to create a python library written in rust. Maybe very few people can achieve what you are trying to create because BPF & rust & python & pyo3 is so rare combination.

And if you have enough time and if you don't mind, please send a PR for this. Perhaps some people in the future would be in trouble with the problem you just fixed.

PR #257, opened 2022-01-14
fpic flag

sebastiaoamaro:
Added fpic flag to the build of "bpf-sys", related to issue #189.

### DIFF
--- a/bpf-sys/build.rs
+++ b/bpf-sys/build.rs
@@ -48,7 +48,7 @@ fn main() {
     // 21.04, Alpine 3.14 also works fine with it
     if !Command::new("make")
         .args(format!("-C libbpf/src BUILD_STATIC_ONLY=1 OBJDIR={out_dir}/libbpf DESTDIR={out_dir} INCLUDEDIR= LIBDIR= UAPIDIR=", out_dir=env::var("OUT_DIR").unwrap()).split(" "))
-        .arg("CFLAGS=-g -O2 -Werror -Wall -fPIE")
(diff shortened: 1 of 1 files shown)


rhdxmr:
@sebastiaoamaro 

Thank you for your contribution.

Please commit your code with `Signed-off-by` trailer.
You can do that by `git commit -s`.
And then re-push your branch with force option.


And it's vague to me what happens if both `-fPIE` and `-fPIC` are specified. So I
think specifying only `-fPIC` is enough.

By compiling with `-fPIC`, the resulting relocatable file can be linked into
PIE executables and shared libraries.

How do you think about replacing `-fPIE` with `-fPIC`? Not just appending
`-fPIC`.

sebastiaoamaro:
I am not an expert in the matter, but after reading the description it makes sense to me to replace -fPIE with -fPIC. I committed the change to this branch https://github.com/sebastiaoamaro/redbpf/tree/fpic-flag

rhdxmr:
@sebastiaoamaro 
Could you do this in your main branch?
1. `git reset HEAD^`. To reset right before the first commit
2. Just remove `-fPIE`
3. `git commit -s`. To commit with signature.
4. `git push -f`. To push main branch forcibly.

I will be able to merge this PR after you do this tasks.

sebastiaoamaro:
Never did a pull request before, Thanks for the patience :) Think it is done now? I hope...

rhdxmr:
Yeah, this is what I expected.
I'm approving this and build-test will start. I hope all build tests succeed.

rhdxmr (approved):

PR #258, opened 2022-01-18
Change default LLVM to 13

rhdxmr:
This PR changes default LLVM version to 13.

It also introduces a protection against LLVM versions mismatch at runtime.
And documentation is updated.


rhdxmr:
oops.. command line options of build-test should be fixed.

rsdy (approved):
Looks ok apart from the build issue.

I think it may be useful to keep the current feature internals as they are, just deprecate the older llvm versions. So, keep the `llvm13` feature, just make it default. This way the build issues will go away automatically, and we can keep rolling the same mechanism forward as needed.

rhdxmr:
I resurrected `llvm13` feature and fixed the build test yml.

PR #262, opened 2022-01-20
Remove loop unroll

rhdxmr:
Since the Linux kernel v5.3, bounded loop has been supported.

### DIFF
--- a/examples/example-userspace/examples/bounded-loop.rs
+++ b/examples/example-userspace/examples/bounded-loop.rs
@@ -1,0 +1,31 @@
+/// This example shows disassemble of bounded_loop probe file.
+///
+/// If `cargo bpf build` is executed, the resulting BPF bytecode is relatively
+/// small. And you can see that the disassembled code is short and you can find
+/// bounded loop code in there. Since the Linux kernel v5.3, the bounded loop
+/// is permitted by the BPF verifier. So the default behavior of `cargo bpf
+/// build` is to preserve the bounded loop if the loop iteration count is big
+/// enough.
+///
+/// If `cargo bpf build --force-loop-unroll` is executed, the resulting BPF
+/// bytecode does not contain loop code. Instead, it contains fully unrolled
+/// loop. So you can see the disassembled code is pretty long.
+use std::path::PathBuf;
+use std::process::Command;
+
+fn main() {
+    println!("Run llvm-objdump --disassemble");
+    println!("========");
+    let _ = Command::new("llvm-objdump")
+        .arg("--disassemble")
+        .arg(probe_path().into_os_string().into_string().unwrap())
+        .status();
+}
+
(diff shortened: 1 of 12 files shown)


rhdxmr:
Cc @kevi-sun

kevi-sun:
nice one!
I'm not quite sure whether someone might use old kernel version, to support that, sit might be good to have and option to toggle this action.

kevi-sun (approved):

rsdy:
I agree, it seems sensible to put this behind a `feature` gate, may be on by default?

rhdxmr:
Oh that sounds great!

Currently all loops are unrolled forcibly.
But I think it is not needed anymore since BPF supports bounded loops.

I think it is better to allow bounded loop by default and let users turn on the loop-unroll feature if they want.
Because the Linux kernel v5.3 is out of our build tests.
Now the oldest kernel version is v5.4 of ubuntu 20.04 LTS.

rhdxmr:
Change cargo-bpf not to force loop unrolling.
`cargo bpf build --force-loop-unroll` option is added to enable full loop unrolling.

examples/example-userspace/examples/bounded-loop.rs
```diff
@@ -0,0 +1,31 @@
...
+/// small. And you can see that the disassembled code is short and you can find
+/// bounded loop code in there. Since the Linux kernel v5.3, the bounded loop
+/// is permitted by the BPF verifier. So the default behavior of `cargo bpf
+/// build` is to preserve the bounded loop if the loop iteration count is big
+/// enough.
+///
+/// If `cargo bpf build --force-loop-unroll` is executed, the resulting BPF
+/// bytecode does not contain loop code. Instead, it contains fully unrolled
+/// loop. So you can see the disassembled code is pretty long.
+use std::path::PathBuf;
```
kevi-sun, line 13: good

kevi-sun (approved):

rhdxmr:
@rsdy 
I added an option `--force-loop-unroll` to `cargo bpf` command.
what do you think about this?

rsdy (approved):

examples/example-userspace/examples/tc-map-share.rs
```diff
@@ -27,7 +27,7 @@ async fn main() {
         .with_max_level(Level::TRACE)
         .finish();
     tracing::subscriber::set_global_default(subscriber).unwrap();
-    if unsafe { libc::getuid() != 0 } {
+    if unsafe { libc::geteuid() != 0 } {
```
rsdy, line 30: Just to note that the `geteuid()` changes are not strictly related to this patchset.

rhdxmr: That's true. It would be better this modification is sent via separate PR.

PR #264, opened 2022-01-21
add p0f as example

kevi-sun:
add p0f extract as one example, while it can only support 5 TCP options and more options would lead to BPF app too complex that failed the Linux BPF verifier.

hope it could be simplified and pass the Linux verifier.

Signed-off-by: Kevin Sun <ksun@darwinium.com>

### DIFF
--- a/examples/example-userspace/examples/p0f.rs
+++ b/examples/example-userspace/examples/p0f.rs
@@ -1,0 +1,49 @@
+// for details about the p0f, please refer to https://github.com/p0f/p0f
+
+use futures::stream::StreamExt;
+use probes::p0f::TcpSignature;
+use redbpf::{load::Loader, xdp};
+
+fn probe_code() -> &'static [u8] {
+    include_bytes!(concat!(env!("OUT_DIR"), "/target/bpf/programs/p0f/p0f.elf"))
+}
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> std::result::Result<(), String> {
+    let xdp_mode = xdp::Flags::DrvMode;
+    let interfaces: Vec<String> = vec!["eth0".to_string()];
+
+    let mut loaded = Loader::load(probe_code()).map_err(|err| format!("{:?}", err))?;
+
+    for interface in &interfaces {
+        println!("Attach p0f on interface: {} with mode {:?}", interface, xdp_mode);
+        for prog in loaded.xdps_mut() {
+            prog.attach_xdp(interface, xdp_mode)
+                .map_err(|err| format!("{:?}", err))?;
+        }
+    }
+
+    let _ = tokio::spawn(async move {
+        while let Some((name, events)) = loaded.events.next().await {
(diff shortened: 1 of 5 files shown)


examples/example-userspace/examples/p0f.rs
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    let _ = tokio::spawn(async move {
+        while let Some((name, events)) = loaded.events.next().await {
+            for event in events {
+                match name.as_str() {
+                    "tcp_signatures" => {
+                        let tcp_sig =
+                            unsafe { std::ptr::read(event.as_ptr() as *const TcpSignature) };
```
rhdxmr, line 32: I think `ptr::read_unaligned` is proper here since the result of `event.as_ptr()` is not guaranteed to be aligned at the `mem::align_of::<TcpSignature>()`

rhdxmr (changes requested):
This example program is brilliant! :+1: 


Could you change the commit message to contain `Signed-off-by` trailer?
Since `redbpf` project conforms Developer Certificate Origin (DCO),
I ask you to do so. So please follow these steps below:
1. Rebase interactively to edit commit messages
2. Append `Singed-off-by` trailer to each commits
3. Push your branch forcibly.

You can append the trailer by your hand or you can make use of `git commit -s` option to do it.

examples/example-userspace/examples/p0f.rs
kevi-sun: @rhdxmr sorry for delayed reply and thanks a lot for your review, I've fixed this and added the "signed-off-by". if you got available time, could you please help to check whether it is possible to simplify it and increase the tcp option count, which is controlled by 

```rust
for _ in 0..p0f::MAX_TCP_OPT - 6 {
```

it is 5 at most and it would be nice to support more

rhdxmr: If the loop iteration count is increased, it is failed to load this BPF program with this BPF verifier log..
```
BPF program is too large. Processed 1000001 insn
processed 1000001 insns (limit 1000000) max_states_per_insn 23 total_states 88465 peak_states 1892 mark_read 18
```
The BPF bytecode of p0f only contains about 2k instructions, so the 1M limit problem is not directly caused by the number of instructions generated by compiler. The thing is the number of executed instructions at runtime.

So I think this problem can not be solved by bounded-loop or BPF-to-BPF call that just reduce the size of BPF bytecode.


This link suggests to use tail call.
https://stackoverflow.com/a/70848246/12851147
But I am not sure.

Although this might solve the problem, a big modification should be involved. So I think it would be better to deal the improvement by another PR if it is possible.

rhdxmr:
I am approving this PR and initiating the build tests.
It will take few hours to finish. After that, I will revisit this PR and complete the merge process.
Thanks @kevi-sun

rhdxmr (approved):

PR #273, opened 2022-02-01
redbpf-probes: add support for XSKMAP eBPF Maps

FelixMcFelix:
This commit adds support for XskMap datastructures (BPF_MAP_TYPE_XSKMAP),
allowing the use of redbpf-generated binaries in AF_XDP programs.

This mostly repurposes the existing `DevMap` code as the redirect logic
is mostly identical.

Signed-off-by: Kyle Simpson <kyleandrew.simpson@gmail.com>

### DIFF
--- a/redbpf-macros/src/lib.rs
+++ b/redbpf-macros/src/lib.rs
@@ -263,7 +263,7 @@ pub fn map(attrs: TokenStream, item: TokenStream) -> TokenStream {
             } else {
                 // without generic types
                 match map_type_name.as_str() {
-                    "StackTrace" | "SockMap" | "ProgramArray" | "DevMap" => {}
+                    "StackTrace" | "SockMap" | "ProgramArray" | "DevMap" | "XskMap" => {}
                     _ => {
                         panic!("unknown map type name: {}", map_type_name);
                     }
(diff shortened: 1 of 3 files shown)


FelixMcFelix:
As an additional thought, the redirect logic could be exposed via a trait, or the implementation could just be broken out to its own function taking a `bpf_map_dev`. I'm unsure what you'd prefer.

In testing, I could only get libbpf to accept generated binaries if I tagged the map as follows:
```rust
#[map(link_section = "maps")]
static mut xsks_map: XskMap = XskMap::with_max_entries(QUEUE_LEN);
```
I'm not sure how this scales to support more maps. A test program seemed to load fine (and correctly steered packets) in spite of illegal identifiers in BTF, although I'm not sure the probe renaming (i.e., "outer_x") is documented anywhere.

rhdxmr:
> As an additional thought, the redirect logic could be exposed via a trait, or the implementation could just be broken out to its own function taking a `bpf_map_dev`. I'm unsure what you'd prefer.

Likewise, I hope to see a design for generic redirect logic. But I didn't find a better solution yet.
I'm expecting some brilliant suggestion will be brought to the table. :slightly_smiling_face: 


> I'm not sure how this scales to support more maps.
Yes, defining multiple maps in `maps` section is supported.

rhdxmr:
This PR looks good to me so I am approving this PR and allowing to run build tests.
The test takes few hours since github action only provides low computing power..
The actual merging process will be done after the tests finish.

rhdxmr (approved):

PR #275, opened 2022-02-05
Check if tc_action program contains relocation for another section

rhdxmr:
`tc` utility does not support relocation for `.rodata` section.
So if `map.get(&42)` is written in the probe, an error occurs when `tc` loads
the probe.
This PR prevents users from writing invalid code like `map.get(&42)` by raising
a compile error with a descriptive error message.


rsdy (approved):

rsdy:
I think this is a good idea. Lgtm.

rhdxmr:
Okay then I am going to merge this PR.
Thanks for your review.

PR #278, opened 2022-02-07
Use libbpf-sys

kriomant:


redbpf/src/lib.rs
```diff
@@ -502,7 +502,7 @@ impl Program {
     }
 
     fn with_btf(kind: &str, name: &str, code: &[u8], btf: &BTF) -> Result<Program> {
-        let code = zero::read_array(code).to_vec();
+        let code = unsafe { zero::read_array_unsafe(code) }.to_vec();
```
rsdy, line 505: What's the motivation behind dropping checks during this step?

rsdy (commented):
Thanks so much for this, it looks great.

Apart from the non-trivial change of changing parsing code to `unsafe`, my only other ask is to re-push the contents of the patchset to please the DCO bot.

You can see more info about that process [here](https://github.com/foniod/project/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)

redbpf/src/lib.rs
kriomant: `zero` crate allows to use safe methods only for structs which implements (unsafe) trait `Pod`. `bpf-sys` crate implemented this trait for some bindings structs. But now it can't, because both struct and trait are from foreign creates. So `unsafe` moved from `impl` to call site.

rsdy: Ah, I forgot about this! Makes sense.

rsdy:
Looking at the CI, we'll need to bump the build images to successfully build this this branch.

rsdy:
I am very confused why the CI doesn't pass on this with the exception of 20.04. Will need to fire up a bunch of Dockers to dig deeper.

rsdy:
@rhdxmr CI seems to fail with an error saying `Feb 13 13:44:42.856 DEBUG build_script_build: Generating bindings with BTF of vmlinux`. Do you think you could look into this?

On aarch64 the error seems to be a bit more straightforward

```
 error[E0412]: cannot find type `__gnuc_va_list` in module `super`
   --> bpf-sys/src/type_gen.rs:273:54
    |
273 |     #[cfg(not(target_env = "musl"))] va_list: super::__gnuc_va_list,
    |                                                      ^^^^^^^^^^^^^^ not found in `super`
```

rhdxmr:
@rsdy 
All problems are solved. I think it is okay now.
I double checked all modifications but it would be better if you check for the last time.

PR #298, opened 2022-03-05
Allow users to define own BPF maps compatible with map! macro

kriomant:


kriomant:
I just thought that there is alternative to generating TC-like metadata for all map types. Since user already explicitly states he wants map to be TC-compatible (by using `TcHashMap` instead of `HashMap`), we may force him to use specific macro for them, like `#[tc_map]` or `#map(tc_compatible)`. Important part is that matching between map type and macro will be checked at compile-time.

rsdy:
I like this a lot. Much cleaner than before. Thanks!

rsdy (approved):

PR #303, opened 2022-03-23
Add ability to use interface index for attaching/detaching

kriomant:
In some cases there is already interface index and user is forced
to convert it to interface name just for RedBPF to immediately
resolve it back to interface index.
Moreover, interface indices are stable, while name of interface
may change on the fly.


rsdy (approved):
LGTM

rsdy (approved):

PR #307, opened 2022-04-03
Add support for USDT semaphores

nbaksalyar:
[USDT probes](https://leezhenghui.github.io/linux/2019/03/05/exploring-usdt-on-linux.html) can be used to define static trace points which can provide helpful structured debug info. Due to performance considerations, these probes can be disabled by default through a mechanism called [_semaphores_](https://leezhenghui.github.io/linux/2019/03/05/exploring-usdt-on-linux.html#heading-sample-w-semaphore-support) which are basically pre-defined memory locations which need to be rewritten to enable probes.

This can be done on a per-process basis by poking a given memory address but you need to know a PID beforehand to do so.
Starting from version 4.20, the Linux kernel provides a mechanism to automatically override semaphore addresses when you setup a uprobe.

This PR implements an optional argument for uprobes which can be used to set a semaphore address to pass on to the kernel. See a similar feature implemented in BCC: https://github.com/iovisor/bcc/pull/3135

I'm wary of changing public APIs, so this PR isn't a complete implementation but rather a start of a discussion and a PoC.

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -60,9 +60,8 @@ use goblin::elf::{reloc::RelocSection, section_header as hdr, Elf, SectionHeader
 use libbpf_sys::{
     bpf_create_map_attr, bpf_create_map_xattr, bpf_insn, bpf_iter_create, bpf_link_create,
     bpf_load_program_xattr, bpf_map_def, bpf_map_info, bpf_prog_type, BPF_ANY, BPF_MAP_TYPE_ARRAY,
-    BPF_MAP_TYPE_HASH, BPF_MAP_TYPE_LRU_HASH, BPF_MAP_TYPE_LRU_PERCPU_HASH,
+    BPF_MAP_TYPE_HASH, BPF_MAP_TYPE_LPM_TRIE, BPF_MAP_TYPE_LRU_HASH, BPF_MAP_TYPE_LRU_PERCPU_HASH,
     BPF_MAP_TYPE_PERCPU_ARRAY, BPF_MAP_TYPE_PERCPU_HASH, BPF_MAP_TYPE_PERF_EVENT_ARRAY,
-    BPF_MAP_TYPE_LPM_TRIE,
     BPF_SK_LOOKUP, BPF_SK_SKB_STREAM_PARSER, BPF_SK_SKB_STREAM_VERDICT, BPF_TRACE_ITER,
 };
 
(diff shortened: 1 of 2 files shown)


rsdy (approved):

rsdy:
It would be helpful to gracefully upgarde the API, ie. introduce a new function that deals with this use case specifically. Also, I'll need to double check if `examples` are actually built with the CI, since I think there is at least one for uprobes.

rsdy:
It would be helpful to gracefully upgarde the API, ie. introduce a new function `attach_uprobe_with_semaphore` that deals with this use case specifically. Also, I'll need to double check if `examples` are actually built with the CI, since I think there is at least one for uprobes.

nbaksalyar:
> I'll need to double check if examples are actually built with the CI, since I think there is at least one for uprobes.

Would you mind if I added an example for USDT + semaphores usage too? (This would likely require introducing a new dev dependency.)

rsdy:
@nbaksalyar Please do! It would be great to expand the coverage.

nbaksalyar:
I updated this change to be non-breaking. As for examples/tests, I realised that they will require an expanded support for USDT probes - e.g., a helper function like `attach_usdt(...)` which will automatically find available probes and semaphore addresses, similarly to what `attach_uprobe` does for uprobes with `ElfSymbols`. So I would prefer to implement examples and tests in a separate PR if that's OK.

I also haven't found any Rust libraries that read USDT probes so this needs to be addressed first. I think it makes sense to implement this logic in a separate crate and include it as a dependency for redbpf. This way, it can be used even outside the context of eBPF. I can include it as a part of the `redbpf` workspace or maintain it separately if that's preferred.

Semaphores are also mostly exclusive to USDTs so arguably we might not even need a separate `attach_uprobe_with_semaphore` function -- i.e., this argument can be moved to `attach_usdt` -- but I think it gives a bit more flexibility since a user might prefer to source probe/semaphores addresses from place other than ELF headers.

Either case, this PR should be good to be accepted or rejected for now. :)

rsdy (approved):

rsdy:
@nbaksalyar Happy for examples to come separately.

If you wish to maintain the USDT library, I'm happy for you to maintain it on your own account, or as part of the `foniod` org, up to you. I'd prefer to keep it out of the redbpf codebase to keep this repo focused on the core, as it's complex enough.

PR #318, opened 2022-04-14
Update RingBufs API - allow referring individual maps

nbaksalyar:
As was discussed in #251, the current ringbufs API has a disadvantage: in order to handle different maps in different async tasks, a user needs to redirect events to e.g. mpsc queues to handle them separately.

This change simplifies the API and implementation by allowing to refer to BPF maps individually through a corresponding key in a `HashMap`.

### DIFF
--- a/redbpf/src/load/loader.rs
+++ b/redbpf/src/load/loader.rs
@@ -7,6 +7,7 @@
 
 use futures::channel::mpsc;
 use futures::prelude::*;
+use std::collections::HashMap;
 use std::convert::AsRef;
 use std::fs;
 use std::io;
@@ -44,6 +45,7 @@ impl Loader {
 
         let online_cpus = cpus::get_online().unwrap();
         let (sender, receiver) = mpsc::unbounded();
+        let mut ringbufs = HashMap::new();
 
         // bpf_map_type_BPF_MAP_TYPE_PERF_EVENT_ARRAY = 4
         for m in module.maps.iter_mut().filter(|m| m.kind == 4) {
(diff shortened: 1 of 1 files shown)


redbpf/src/load/loader.rs
```diff
@@ -108,6 +106,25 @@ pub struct Loaded {
...
+    /// use redbpf::load::Loader;
+    /// # async {
+    /// let mut loader = Loader::load_file(&Path::new("probe.elf")).unwrap();
+    /// let mut ringbuf = loader.ringbufs.get("ring_buffer");
+    /// while let Some(event) = ringbuf.next().await {
+    ///     // ...
+    /// }
+    /// # };
+    /// ```
+    pub ringbufs: HashMap<String, RingBufMessageStream>,
```
nbaksalyar, line 127: I have a reservation about how this is exposed in the API: there's a discrepancy in how ringbuf and perfmap events are handled.

I think it would make sense to use the same approach for `PerfMap`s too? This would be a breaking change though.

rsdy: As per our agreement with @rhdxmr we are going to bump major version for the next release due to the change in default LLVM. The time to break APIs is now.

nbaksalyar: Sounds good! In this case, I'd advocate for adding a changelog - created an issue #319

rsdy:
@nbaksalyar What's your plan here? Do you want to get this PR merged, or you'd rather amend this PR with the API changes you mentioned above?

nbaksalyar:
> What's your plan here? Do you want to get this PR merged, or you'd rather amend this PR with the API changes you mentioned above?

Yes, I think it's better to address these within this PR because the API will change anyway. It is a bit non-trivial though. I'm trying to figure out how to make it universal and not a chore to use. Most likely it'll require using boxed traits or enums for different map types. I'll prioritise finishing this PR.

Issue #320, opened 2022-04-24
bpf-sys, handle when `/proc/version_signature` is empty


QuiiBz:
Hello, I tried to compile my BPF program on a system that's running proxmox, but strangely the `/proc/version_signature` is empty - it seems like it's due to `/boot/config-5.13.19-4-pve` having an empty `CONFIG_VERSION_SIGNATURE=""`.

Here's the exact error when running my loader:

```bash
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/redbpf-2.3.0/src/lib.rs:2765:54
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
which refers to [here](https://github.com/foniod/redbpf/blob/main/bpf-sys/src/uname.rs#L28). As I can see, the `parse_version_signature` doesn't handle the fact that `/proc/version_signature` can be empty, and so return a `None`.

Would you like me to create a PR to fix this issue, or do you need any more information? Thanks.

nbaksalyar:
That's a good catch. I think it would make sense to get rid of this `unwrap` in `fn get_version` in addition to handling the case of an empty string in `parse_version_signature`.

`get_version` is used [only in `ModuleBuilder::parse`](https://github.com/foniod/redbpf/blob/c864a69e4cb3beac1cdf1d01b07cc3ae89b17368/redbpf/src/lib.rs#L1523), so this line can be easily replaced with something like `get_version(&content).ok_or(Error::InvalidVersion)`.

QuiiBz:
Would I still be able to run the loader with your proposition? Looks like the version is used at other spots.

I'm thinking about checking in [`get_kernel_internal_version `](https://github.com/foniod/redbpf/blob/main/bpf-sys/src/uname.rs#L28) if the file is empty or not before calling `parse_version_signature`, to fallback to `uname()`. I think it's doesn't make a lot of sense to parse the version from a file that is empty 🤔 .

nbaksalyar:
> checking in [get_kernel_internal_version ](https://github.com/foniod/redbpf/blob/main/bpf-sys/src/uname.rs#L28) if the file is empty or not before calling parse_version_signature

Sure, but `fs::read_to_string` should take care of that for you because there's no data associated with inode - so checking if a file is empty would be superfluous. So I would just do something like

```rust
let mut version = None;
if let Ok(version) = fs::read_to_string("/proc/version_signature") {
[...]

PR #322, opened 2022-04-24
fix(bpf-sys): use uname when /proc/version_signature is empty

QuiiBz:
Following the discussion at https://github.com/foniod/redbpf/issues/320,  this PR fixes an error when `/proc/version_signature` exists but is empty.

### DIFF
--- a/bpf-sys/src/uname.rs
+++ b/bpf-sys/src/uname.rs
@@ -25,11 +25,11 @@ pub fn uname() -> Result<::libc::utsname, ()> {
 
 #[inline]
 pub fn get_kernel_internal_version() -> Option<u32> {
-    let mut version = None;
-
-    if let Ok(version_signature) = fs::read_to_string("/proc/version_signature") {
-        version = parse_version_signature(&version_signature.trim());
-    }
+    let version = if let Ok(version_signature) = fs::read_to_string("/proc/version_signature") {
+        parse_version_signature(&version_signature.trim())
+    } else {
+        None
+    };
 
     let final_version = match version {
         Some(version) => version,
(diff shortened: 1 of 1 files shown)


bpf-sys/src/uname.rs
```diff
@@ -25,13 +25,18 @@ pub fn uname() -> Result<::libc::utsname, ()> {
...
 pub fn get_kernel_internal_version() -> Option<u32> {
-    let version = if let Ok(version) = fs::read_to_string("/proc/version_signature") {
-        parse_version_signature(&version.trim())?
-    } else {
-        to_str(&uname().ok()?.release).into()
+    let mut version = None;
+
+    if let Ok(version_signature) = fs::read_to_string("/proc/version_signature") {
+        version = parse_version_signature(&version_signature.trim());
+    }
```
nbaksalyar, line 32: Since you're using a separate var for `final_version` (which is a cleaner way!), I think `version` doesn't have to be a mutable var:
```rust
    let version = if let Ok(version_signature) = fs::read_to_string("/proc/version_signature") {
        parse_version_signature(&version_signature.trim())
    } else {
        None
    };
```
Or, even simpler:
```suggestion
    let version = fs::read_to_string("/proc/version_signature").ok()
        .and_then(|version_signature| parse_version_signature(&version_signature.trim()));
```

nbaksalyar (commented):
Thanks for the PR! I have one minor suggestion, otherwise LGTM. :+1:

QuiiBz:
Thanks! I applied your first suggestion, feel like it's a bit easier to read.

nbaksalyar (approved):
LGTM :+1: Thank you!

PR #326, opened 2022-05-03
Parse ___btf_map_<map_sym_name> structure to get key/value type id

### LINKED ISSUE #325: BTF illegal structure name
Hello,

I've recently been playing around with RedBPF and had cool results - big thanks for this lib!

Originally I worked with RedBPF v2.3.0, but because I was especially excited about some new features (like the ring buffer) you implemented after the latest release I decided to go bleeding edge and work right off the master branch.

I'm obviously aware that this means a much less stable version, but that's fine for me. However, whenever I build & load my probes with any recent master version (tested commit c864a69e4cb3beac1cdf1d01b07cc3ae89b17368 and 975d87ef92d0ba4a8ae604ed5a6287a622701662) I get the following error messages on startup:

```
ERROR redbpf::btf: illegal structure name: ____btf_map_VARIABLE
ERROR redbpf::btf: illegal structure name: ____btf_map_OUTPUT
ERROR redbpf::btf: illegal structure name: ____btf_map_CONN_MAP
ERROR redbpf::btf: illegal structure name: ____btf_map_PROBES
...
```

This error seems to appear for every single map I've got in my probes - one per map. The BTF names match my map names.

Funnily enough, all of my maps load up and work just fine though. It's just a lot of nagging error messages on startup. It appears as if cargo-bpf compiles some structures into the elf with names the loader doesn't expect? Not sure what's wrong.

My cargo-bpf is on the same bleeding edge version as my redBPF libs, so I don't think it's a version mismatch issue?

This does not happen when I roll back to redBPF v2.3.0, but I really like the new features ❤️

rhdxmr:

This resolves #325

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -60,9 +60,17 @@ use goblin::elf::{reloc::RelocSection, section_header as hdr, Elf, SectionHeader
 use libbpf_sys::{
     bpf_create_map_attr, bpf_create_map_xattr, bpf_insn, bpf_iter_create, bpf_link_create,
     bpf_load_program_xattr, bpf_map_def, bpf_map_info, bpf_prog_type, BPF_ANY, BPF_MAP_TYPE_ARRAY,
-    BPF_MAP_TYPE_HASH, BPF_MAP_TYPE_LPM_TRIE, BPF_MAP_TYPE_LRU_HASH, BPF_MAP_TYPE_LRU_PERCPU_HASH,
-    BPF_MAP_TYPE_PERCPU_ARRAY, BPF_MAP_TYPE_PERCPU_HASH, BPF_MAP_TYPE_PERF_EVENT_ARRAY,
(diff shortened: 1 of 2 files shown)


redbpf/src/lib.rs
```diff
@@ -1893,6 +1902,19 @@ impl Map {
                         }
                     }
                 }
+            } else if let Some(ENOTSUPP) = io::Error::last_os_error().raw_os_error() {
+                debug!(
+                    "Since kernel does not support BTF for {}, BTF is ignored when creating the map `{}'", map_type_name(config.type_), name
```
rsdy, line 1907: "The kernel does not support BTF for {}. Ignoring BTF for map `{}`." Would be better style.

rsdy (approved):

rhdxmr:
I applied your suggestion and repushed the branch.
I'm merging this PR.
Thanks.

PR #331, opened 2022-05-31
Improve support for tracepoints

GermanCoding:
This adds a macro to aid in creating tracepoints, adds various documentation & an example for tracepoints and fixes a bug that prevented tracepoints from attaching.

This is essentially a light version of #124 which does not attempt to generate the tracepoint structs automatically. Instead users are expected to do this themselves. This is not entirely ideal, but given the issues #124 had with this I believe this is the simplest way to go.

This also does not change existing API functions and as such remains largely compatible.


rsdy (approved):
This looks really good! Thanks!

PR #347, opened 2022-09-21
feat: add XDP example for UDP port redirect

shaneutt:
The purpose of this PR is to add an XDP example of redirecting ports for UDP traffic.

This is not complete, and I'm not certain that this is the best or most idiomatic way to do this yet so I'm leaving this in draft for the moment in hopes of getting some input/suggestions from the maintainers.

### DIFF
--- a/redbpf-probes/src/net.rs
+++ b/redbpf-probes/src/net.rs
@@ -25,7 +25,10 @@ pub enum Transport {
     TCP(*const tcphdr),
     UDP(*const udphdr),
 }
-
+pub enum Ip {
+    IPv4(*const iphdr),
+    IPv6(*const ipv6hdr),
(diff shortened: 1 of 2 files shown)


rsdy:
This looks like a passable approach to me, if it works, that's even better. Maybe @rhdxmr has some suggestions on improving it? Regardless, I'd be happy to merge it if you find this helpful!

shaneutt:
> This looks like a passable approach to me, if it works, that's even better.

Thank you. I have tested this locally and it works. **However** I _have yet to get it working with a redbpf userspace loader given the examples available in this repository_. I was instead able to load and test it using the normal `bpftool` approach. This is why I didn't consider it complete yet.

PR #357, opened 2022-12-28
redbpf: compute symbol offset rather than address

### LINKED ISSUE #356: sym_offset calculation in attach uprobe fails to account to address differences
BPF probes want the offset of a symbol in the object file. In shared object and executable files, at least, the current logic seems to return the virtual address of the symbol as opposed to the offset in the object file. 

The `st_value` for a `goblin::elf:sym::Sym` with `st_info` of `STT_FUNC` (`2`) refers to the virtual memory address. In order to compute the offset, one needs to find the offset by translating it to be relative to the address of the section containing the symbol and then adding the offset of the section in the file. 

The offending code is:

https://github.com/foniod/redbpf/blob/55331987dc31a26feb192e4a668bf9bfc336e1e0/redbpf/src/lib.rs#L965

In some cases, the address of the symbol is its offset, but certainly not in all cases. 

See: https://github.com/iovisor/bpftrace/issues/1763#issuecomment-1238935968

ajwerner:
Fixes #356 

BPF expects offsets in the object files rather than virtual memory addresses. In some cases, the virtual memory offset of the .text section is 0, in which case the correction in this change is a no-op. If the .text section has a non-zero address offset, the previous logic for determining the symbol offset was wrong.

### DIFF
--- a/redbpf/src/lib.rs
+++ b/redbpf/src/lib.rs
@@ -958,11 +958,18 @@ impl UProbe {
         };
         let sym_offset = if let Some(fn_name) = fn_name {
             let data = fs::read(&path)?;
-            let parser = ElfSymbols::parse(&data)?;
-            parser
+            let elf = Elf::parse(&data)?;
+            let sym = ElfSymbols::new(&elf)
                 .resolve(fn_name)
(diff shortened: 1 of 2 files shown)


ajwerner:
Preempted by https://github.com/foniod/redbpf/pull/308.

